PR #541, opened 2020-10-04
Ap refactoring

Tirokk:
> [Mastercuber](https://github.com/Mastercuber) **Authored by [Mastercuber](https://github.com/Mastercuber)**
_<time datetime="2019-05-03T18:47:15Z" title="Friday, May 3rd 2019, 8:47:15 pm +02:00">May 3, 2019</time>_
_Closed <time datetime="2019-08-01T10:54:02Z" title="Thursday, August 1st 2019, 12:54:02 pm +02:00">Aug 1, 2019</time>_
---

## Pullrequest
This Pullrequest refactors some files of the ActivityPub component. In addition it add the token on demand funtionality, fixes the problem with using the wrong id and also fixes the false negative creation of a post object test.

### Issues
- fixes #221
- fixes #398 
- closes #413

### Checklist
- [X] None

### How2Test
- [X] None

### Todo
- [X] None

PR #823, opened 2020-10-05
Whitelist queries/mutations, fix tests

### LINKED ISSUE #343: 🐛 Whitelist all auto-generated Mutations/Queries
> [ulfgebhardt](https://github.com/ulfgebhardt) **Authored by [ulfgebhardt](https://github.com/ulfgebhardt)**
_<time datetime="2019-04-01T15:09:52Z" title="Monday, April 1st 2019, 5:09:52 pm +02:00">Apr 1, 2019</time>_
_Closed <time datetime="2019-06-14T11:22:01Z" title="Friday, June 14th 2019, 1:22:01 pm +02:00">Jun 14, 2019</time>_
---

**Describe the bug**
Disable all Mutations/Queries via middleware and recheck which can be enabled under which circumstance.

We use [neo4j-graphql-js](https://github.com/neo4j-graphql/neo4j-graphql-js) as a query builder. None of the generated mutations/queries has authentication/authorization by default. Therefore we need to white-list only those mutations that we need and which are tested on security we before migrating our [production server](https://alpha.human-connection.org/) to the new nitro code base.

**To Reproduce**
1. Go to 'http://localhost:4000'
2. Create e.g. a comment:
```graphql
mutation {
  CreateComment(content: "I am a comment") {
    id
  }
}
```
3. Click submit
3. You just created a comment without an author.

**Expected behavior**
No private data should be exposed, ie. emails of users. No disabled or deleted content should be visible. No data should be created without authorship of the current user.

Tirokk:
_<time datetime="2019-06-13T17:51:00Z" title="Thursday, June 13th 2019, 7:51:00 pm +02:00">Jun 13, 2019</time>_
_Merged <time datetime="2019-06-14T11:22:01Z" title="Friday, June 14th 2019, 1:22:01 pm +02:00">Jun 14, 2019</time>_
---

## 🍰 Pullrequest

This Pull Request throw an authorization error on all Queries/Mutations by default, unless we explicitly specifiy how authorization should behave.
The types on the other hand will be allowed by default. The types are checked *after* the resolver of the query/mutation has been called, the db requests have resolved and the response is on its way out. We e.g. have some extra checks on the `email`, `password`, and `privateKey` field of the user, so those attributes can never be read.

@Mastercuber there are two failing cucumber scenarios now. I disabled these cucumber scenarios for reasons I explained in the commit message.

### Issues

- close #343 

### Todo

- [ ] Refactor ActivityPub middleware (will be done when we have a look into #541)


backend/src/middleware/permissionsMiddleware.js
```diff
@@ -48,6 +54,13 @@ const belongsToMe = rule({
   return Boolean(notification)
 })
 
+/* TODO: decide if we want to remove this check: the check
+ * `onlyEnabledContent` throws authorization errors only if you have
+ * arguments for `disabled` or `deleted` assuming these are filter
+ * parameters. Soft-delete middleware obfuscates data on its way out
+ * anyways. Furthermore, `neo4j-graphql-js` offers many ways to filter for
+ * data so I believe, this is not a good check anyways.
+ */
```
Tirokk, line 63: > [roschaefer](https://github.com/roschaefer) **Authored by [roschaefer](https://github.com/roschaefer)**
_<time datetime="2019-06-14T09:39:52Z" title="Friday, June 14th 2019, 11:39:52 am +02:00">Jun 14, 2019</time>_

---
> **Outdated (history rewrite)** - original diff
---
```diff
@@ -48,6 +54,13 @@ const belongsToMe = rule({
   return Boolean(notification)
 })
 
+/* TODO: decide if we want to remove this check: the check
+ * `onlyEnabledContent` throws authorization errors only if you have
+ * arguments for `disabled` or `deleted` assuming these are filter
+ * parameters. Soft-delete middleware obfuscates data on its way out
+ * anyways. Furthermore, `neo4j-graphql-js` offers many ways to filter for
+ * data so I believe, this is not a good check anyways.
+ */
```


What do we do about `enabledContent`? Shall I remove it?

Tirokk:
_<time datetime="2019-06-14T10:42:43Z" title="Friday, June 14th 2019, 12:42:43 pm +02:00">Jun 14, 2019</time>_

---


@ogerly this PR is not related to images :thinking:

backend/src/middleware/permissionsMiddleware.js
Tirokk: > [ulfgebhardt](https://github.com/ulfgebhardt) **Authored by [ulfgebhardt](https://github.com/ulfgebhardt)**
_<time datetime="2019-06-14T11:14:12Z" title="Friday, June 14th 2019, 1:14:12 pm +02:00">Jun 14, 2019</time>_

---


This clearly depends on who does the filtering - if its the server filtering stuff its fine to remove the check. If the filter is defined by the client and not enforced by the server, we need to make sure only content is delivered which the requesting user is allowed to see.

Tirokk: > [roschaefer](https://github.com/roschaefer) **Authored by [roschaefer](https://github.com/roschaefer)**
_<time datetime="2019-06-14T11:21:32Z" title="Friday, June 14th 2019, 1:21:32 pm +02:00">Jun 14, 2019</time>_

---


The 100% security that the users see only what he is allowed to see is **not** handled by this check.

The check does sth. else: If you filter for posts that are deleted **only**, you will get an authentication error right away in case you are not an administrator. But if we remove this check, then you would get a result set with all deleted posts obfuscated. Same effect but different behaviour.

The obfuscation is the 100% security, because it's implemented on the post itself, not the top-level query. So even if you request nested posts (e.g. all posts of a user) you will see the obfuscated data.

Mogge (approved):
the pictures are excited cleanly

Mogge (approved):
Approving this. This is a step into the right direction and its nice to see that all things have been working as expected. Please review my comment on your question!

PR #1188, opened 2020-10-05
_Skipped manually - Creating pull- 88

Tirokk:


Tirokk:
_<time datetime="2019-11-06T15:23:24Z" title="Wednesday, November 6th 2019, 4:23:24 pm +01:00">Nov 6, 2019</time>_

---


could you give more info on this @Tirokk ??
do you mean that a user when editing their post should see a delete icon to delete their existing image?
Or that after a user adds an image their should be a way to remove it before creating/saving a Post?
both?

Tirokk:
> [Tirokk](https://github.com/Tirokk) **Authored by [Tirokk](https://github.com/Tirokk)**
_<time datetime="2019-12-12T08:40:23Z" title="Thursday, December 12th 2019, 9:40:23 am +01:00">Dec 12, 2019</time>_

---


I meant your first guess @mattwr18 .

- A user should see a delete icon to delete their existing image when editing their post.

Your second guess is fulfilled on production already in my eyes:

- After a user adds an image there is a way to remove it before creating/saving a post.
  - There is a deletion icon in the upper right corner.

PR #1282, opened 2020-10-05
Get rid of property warnings

Tirokk:
_<time datetime="2019-08-16T15:06:50Z" title="Friday, August 16th 2019, 5:06:50 pm +02:00">Aug 16, 2019</time>_
_Merged <time datetime="2019-08-19T08:06:10Z" title="Monday, August 19th 2019, 10:06:10 am +02:00">Aug 19, 2019</time>_
---

If we always choose `0` as the default value for `endVal` in case it is
not given (maybe apollo assigns null when the request is in flight) then
just make `0` the default.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/webapp/components/CountTo.vue
+++ b/webapp/components/CountTo.vue
@@ -20,7 +20,7 @@ export default {
   },
   props: {
     startVal: { type: Number, default: 0 },
-    endVal: { type: Number, required: true },
+    endVal: { type: Number, default: 0 },
     duration: { type: Number, default: 3000 },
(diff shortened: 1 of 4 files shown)


Mogge (approved):
@roschaefer I approved this and gonna merge cause it seems like an improvement, but if you say get rid of property warnings, it would be nice if you could upload a screen shot, so the `PR` reviewer (that's me :wink: ) will know what you are talking about. :smile_cat:

PR #1397, opened 2020-10-05
Check there are ids in the badIds array

Tirokk:
_<time datetime="2019-08-26T14:22:10Z" title="Monday, August 26th 2019, 4:22:10 pm +02:00">Aug 26, 2019</time>_
_Merged <time datetime="2019-08-30T12:08:14Z" title="Friday, August 30th 2019, 2:08:14 pm +02:00">Aug 30, 2019</time>_
---

- if there are no ids, we shouldn't add an empty array since it adds unnecessarily to our auto-generated post query and affects greatly performance. see issue #1390

### Issues

- relates #1390... too early to say if it fixes it or not

### Todo

- [ ] Test it further @roschaefer


Tirokk:
_<time datetime="2019-08-26T14:53:10Z" title="Monday, August 26th 2019, 4:53:10 pm +02:00">Aug 26, 2019</time>_

---


Nice improvement @roschaefer... I'll merge when it passes the build

webapp/store/search.js
```diff
@@ -46,8 +46,8 @@ export const actions = {
     await this.app.apolloProvider.defaultClient
       .query({
         query: gql`
-          query findPosts($query: String!) {
-            findPosts(query: $query, limit: 10) {
```
Tirokk, line 50: > [roschaefer](https://github.com/roschaefer) **Authored by [roschaefer](https://github.com/roschaefer)**
_<time datetime="2019-08-27T13:49:08Z" title="Tuesday, August 27th 2019, 3:49:08 pm +02:00">Aug 27, 2019</time>_

---
> **Outdated (history rewrite)** - original diff
---
```diff
@@ -46,8 +46,8 @@ export const actions = {
     await this.app.apolloProvider.defaultClient
       .query({
         query: gql`
-          query findPosts($query: String!) {
-            findPosts(query: $query, limit: 10) {
```


I find `query` a better naming for a string that is used for full-text search.

Tirokk: > [mattwr18](https://github.com/mattwr18) **Authored by [mattwr18](https://github.com/mattwr18)**
_<time datetime="2019-08-27T14:08:52Z" title="Tuesday, August 27th 2019, 4:08:52 pm +02:00">Aug 27, 2019</time>_

---


I could agree with that, but if it's named query it breaks the functionality. not sure why exactly cause I'm not sure I completely understand the flow here, but it seems it passes through the `schema.gql` where we use the magic here https://github.com/Human-Connection/Human-Connection/blob/master/backend/src/schema/types/schema.gql#L10
then it gets passed to the `posts` resolver, which calls the `filterForBlockedUsers`
then cause there are no `badIds` it returns the `params`, otherwise it adds them to the `filter`, but the posts which match the query have already been returned from the database :point_up:

webapp/store/search.js
```diff
@@ -46,8 +46,8 @@ export const actions = {
...
+          query findPosts($query: String!, $filter: _PostFilter) {
+            findPosts(query: $query, limit: 10, filter: $filter) {
```
Tirokk, line 50: > [roschaefer](https://github.com/roschaefer) **Authored by [roschaefer](https://github.com/roschaefer)**
_<time datetime="2019-08-30T12:07:30Z" title="Friday, August 30th 2019, 2:07:30 pm +02:00">Aug 30, 2019</time>_

---
> **Outdated (history rewrite)** - original diff
---
```diff
@@ -46,8 +46,8 @@ export const actions = {
     await this.app.apolloProvider.defaultClient
       .query({
         query: gql`
-          query findPosts($query: String!) {
-            findPosts(query: $query, limit: 10) {
+          query findPosts($query: String!, $filter: _PostFilter) {
+            findPosts(query: $query, limit: 10, filter: $filter) {
```


This is the only required change? Hell yeah.

Mogge (approved):

Mogge (changes requested):
I don't understand why the build server passes. To my understanding, the `Search.feature` should be broken and one of the scenarios for the blocked user feature should be broken, too.

Mogge (changes requested):
I tested this locally and it seems that once you block a user, the search feature is broken. That explains why the build server was passing: The test user just does not have any blocked users on the list, thus the search feature works fine.

Screenshot - 2019-08-29T014356 398

I knew it couldn't be that easy @mattwr18 :wink:

Mogge (approved):
Nice

PR #1630, opened 2020-10-05
Fix bug where about must not be empty string

Tirokk:
_<time datetime="2019-09-18T11:11:10Z" title="Wednesday, September 18th 2019, 1:11:10 pm +02:00">Sep 18, 2019</time>_
_Merged <time datetime="2019-09-18T12:39:46Z" title="Wednesday, September 18th 2019, 2:39:46 pm +02:00">Sep 18, 2019</time>_
---

- add test case to test this use case since it is the default from our
UI
At the moment people cannot sign up unless they add their `About` property, which is not the desired functionality


Mogge (approved):
blank

PR #1751, opened 2020-10-05
fix: Github's security vulnerability warning

Tirokk:
_<time datetime="2019-09-28T13:12:37Z" title="Saturday, September 28th 2019, 3:12:37 pm +02:00">Sep 28, 2019</time>_
_Merged <time datetime="2019-09-30T17:20:01Z" title="Monday, September 30th 2019, 7:20:01 pm +02:00">Sep 30, 2019</time>_
---

We use an unmaintained package called `activitystrea.ms` for the
activity pub middleware. Since the activity pub middleware is dead code
at the moment I suggest that we remove that package and disable all the
code that depends on the package.

When we get back to implement the ActivityPub spec for Human Connection I
would suggest to maintain this package `activitystrea.ms`, update the
dependencies there and re-enable and **test** the code of the activity pub
middleware.

## 🍰 Pullrequest
Here's the security warning:

Screenshot - 2019-09-28T151413 514

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -43,7 +43,6 @@
   "dependencies": {
     "@hapi/joi": "^16.1.2",
     "@sentry/node": "^5.6.2",
-    "activitystrea.ms": "~2.1.3",
     "apollo-cache-inmemory": "~1.6.3",
     "apollo-client": "~2.6.4",
     "apollo-link-context": "~1.0.19",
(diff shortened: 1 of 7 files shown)


Tirokk:
_<time datetime="2019-09-30T17:19:21Z" title="Monday, September 30th 2019, 7:19:21 pm +02:00">Sep 30, 2019</time>_

---


@alina-beck good question. Normally I would do it just how you suggest it: Remove code. Don't comment it.

The reason for commenting code is my slight hope that people feel motivated to get it back in when they see it. I'm not super optimistic, though :laughing: 

If you're up for it: All we need to do is to maintain https://github.com/jasnell/activitystrea.ms a little: Update the lodash version there, fix the tests after the build failures. I think it's a great package, just not very often used.

Mogge (approved):
not pretty but should work 🤷‍♀ 😄 

when you say the whole activitypub middleware is 'dead code' at the moment could we just throw it out / delete it properly and then restore it from older commits if and when we need it again? why is it disabled now, anyway? are we planning to put it back in?

PR #2025, opened 2020-10-05
fix: Don't attempt to save locale if not authenticated

Tirokk:
_<time datetime="2019-10-24T09:12:20Z" title="Thursday, October 24th 2019, 11:12:20 am +02:00">Oct 24, 2019</time>_
_Merged <time datetime="2019-10-24T12:22:09Z" title="Thursday, October 24th 2019, 2:22:09 pm +02:00">Oct 24, 2019</time>_
---

~@ogerly the maintenance page needs just two components of the webapp
folder. For that reason we import only these. If you add imports to
`LocaleSwitch` component, you need to copy those files in the
Dockerfile.~

~Please don't merge.~ Maintenance page should not need `graphql` because it's never sending a backend request.

### Issues

- None

### Todo

- [X] None


Mogge (approved):
Yeah, good job @mattwr18 !!! 🚀💫
And thanks for @ogerly suggestion with the current user !!! 🚀💫

Issue #3779, opened 2020-10-06
🚀 [Feature] Show capslock icon on password fields


Tirokk:
> [<img alt="sushidave" height="40" width="40" align="left" src="https://avatars1.githubusercontent.com/u/38261864?s=88&v=4">](https://github.com/sushidave) **Authored by [sushidave](https://github.com/sushidave)**
_<time datetime="2020-08-05T11:10:36Z" title="Wednesday, August 5th 2020, 1:10:36 pm +02:00">Aug 5, 2020</time>_

---

## :rocket: Feature

A capslock icon should appear in password fields left to the show/hide button if the capslock key is active.

### User Problem

Currently, the UI doesn't support users getting aware of their capslock key being active and that the login or changing a password might fail. This feature:
- Provides a better user experience and satisfaction
- Increases the rate of successful logins and change of passwords
- Reduces system load

### Implementation

- Add a capslock icon to the login form that responds to the capslock key
- Add it to other password fields of the network.

### Design & Layout

![show-capslock-icon-on-login-form](https://user-images.githubusercontent.com/17728384/89404516-6c0af380-d71a-11ea-99fe-34fbad71879e.png)

### Validation

- Better user experience and satisfaction
- Reduced system load

### Additional context

This feature can be implemented separately from #2434 which solves a similar problem.

Tirokk:
> [<img alt="enigmatic241" height="40" width="40" align="left" src="https://avatars1.githubusercontent.com/u/38261864?s=88&v=4">](https://github.com/enigmatic241) **Authored by [enigmatic241](https://github.com/enigmatic241)**
_<time datetime="2020-08-24T05:50:36Z" title="Monday, August 24th 2020, 7:50:36 am +02:00">Aug 24, 2020</time>_

---


@sushidave I want to work on this.

Tirokk:
> [<img alt="sushidave" height="40" width="40" align="left" src="https://avatars1.githubusercontent.com/u/38261864?s=88&v=4">](https://github.com/sushidave) **Authored by [sushidave](https://github.com/sushidave)**
_<time datetime="2020-08-24T13:23:58Z" title="Monday, August 24th 2020, 3:23:58 pm +02:00">Aug 24, 2020</time>_

---


Hello @enigmatic241 and greetings to India! Many thanks for your message and your readiness to work on this issue. I'm asking @Tirokk to provide you with the necessary details. He'll be back to work this week and supporting you if you need any help.

[...]

PR #4031, opened 2020-12-01
feat: 🍰 Password Field Capslock Alert

Brandon-G-Tripp:
## 🍰 Pullrequest

Pull Request creates a Caps Lock enabled warning when user is typing in the password field on login. 

### Issues

- #3779 

There is one small bug. If the capslock is already enabled on page refresh it will temporarily not trigger when the password field is clicked. I do not believe I can resolve this issue fully until the ds-input is migrated and there is access to the change event. 

Should this issue be put on hold until I finish migration? 

### Todo

- [ ] Fix Test Case for changing data visibility of warning
- [ ] Finalize styling of warning

### DIFF
--- a/webapp/components/LoginForm/LoginForm.spec.js
+++ b/webapp/components/LoginForm/LoginForm.spec.js
@@ -63,14 +63,13 @@ describe('LoginForm', () => {
       const wrapper = Wrapper()
 
       it('Caps-lock warning message is hidden on page load', () => {
-        expect(wrapper.find('text.caps-warning').attributes('style')).toEqual('display: none;')
+        expect(wrapper.find('.caps-warning').isVisible()).toBe(false)
       })
 
-      it('shows warning when caps-lock is press', async () => {
-        const formWrapper = wrapper.find({ ref: 'form' })
-        formWrapper.trigger('keydown', {key: 'CapsLock'})
-        // await Vue.nextTick()
-        expect(wrapper.vm.caps).toBe(true)
+      it('shows warning when caps-lock is pressed', async () => {
+        wrapper.setData({ caps: true })
+        await Vue.nextTick()
+        expect(wrapper.find('.caps-warning').isVisible()).toBe(true)
       })
     })
   })
--- a/webapp/components/LoginForm/LoginForm.spec.js
+++ b/webapp/components/LoginForm/LoginForm.spec.js
@@ -1,3 +1,4 @@
+import Vue from 'vue'
 import LoginForm from './LoginForm.vue'
 import Styleguide from '@human-connection/styleguide'
 import Vuex from 'vuex'
@@ -55,6 +56,21 @@ describe('LoginForm', () => {
           email: 'email@example.org',
           password: '1234',
         })
+      })
+    })
+
+    describe('Warning message is visible when caps lock is on', () => {
+      const wrapper = Wrapper()
+
+      it('Caps-lock warning message is hidden on page load', () => {
+        expect(wrapper.find('text.caps-warning').attributes('style')).toEqual('display: none;')
+      })
+
+      it('shows warning when caps-lock is press', async () => {
+        const formWrapper = wrapper.find({ ref: 'form' })
+        formWrapper.trigger('keydown', {key: 'CapsLock'})
+        // await Vue.nextTick()
+        expect(wrapper.vm.caps).toBe(true)
       })
     })
   })
--- a/webapp/components/LoginForm/LoginForm.vue
+++ b/webapp/components/LoginForm/LoginForm.vue
@@ -31,6 +31,7 @@
             name="password"
             type="password"
             ref="passwordInput"
+            @focus="capsLock"
           />
         <nuxt-link to="/password-reset/request">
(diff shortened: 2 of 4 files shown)


Brandon-G-Tripp:
@Mogge I pushed up the CapsLock branch so you could take a look at the styling. I am not sure what exactly it should look like.

webapp/components/LoginForm/LoginForm.spec.js
```diff
@@ -57,5 +58,20 @@ describe('LoginForm', () => {
...
+    describe('Warning message is visible when caps lock is on', () => {
+      const wrapper = Wrapper()
+
+      it('Caps-lock warning message is hidden on page load', () => {
+        expect(wrapper.find('text.caps-warning').attributes('style')).toEqual('display: none;')
+      })
+
+      it('shows warning when caps-lock is press', async () => {
+        const formWrapper = wrapper.find({ ref: 'form' })
+        formWrapper.trigger('keydown', {key: 'CapsLock'})
```
Brandon-G-Tripp, line 71: The one issue I am having with this test case is that I am not able trigger or emit an event to simulate the changing of the data value from false to true.

ulfgebhardt:
Hey  @Brandon-G-Tripp whenever you are ready lets get this merged then. I would just require you to make an issue for the one problem you encountered.

Maybe @Tirokk or @Mogge wanna have another brief look.

webapp/components/LoginForm/LoginForm.spec.js
```diff
@@ -57,5 +58,20 @@ describe('LoginForm', () => {
         })
       })
     })
+
+    describe('Warning message is visible when caps lock is on', () => {
+      const wrapper = Wrapper()
+
+      it('Caps-lock warning message is hidden on page load', () => {
+        expect(wrapper.find('text.caps-warning').attributes('style')).toEqual('display: none;')
```
Mogge, line 66: Better
```js
expect(wrapper.find('.caps-warning').isVisible()).toBe(false)
```

webapp/components/LoginForm/LoginForm.vue
```diff
@@ -20,18 +24,20 @@
...
+            icon="lock"
+            icon-right="question-circle"
+            name="password"
+            type="password"
+            ref="passwordInput"
+          />
         <nuxt-link to="/password-reset/request">
           {{ $t('login.forgotPassword') }}
         </nuxt-link>
+        <text class="caps-warning" v-show="caps" >CAPS LOCK ENABLED!</text>
```
Mogge, line 40: Please localize the text. German would be: `Feststelltaste ist eingerastet`

webapp/components/LoginForm/LoginForm.vue
```diff
@@ -11,7 +11,11 @@
         </a>
       </template>
       <h2 class="title">{{ $t('login.login') }}</h2>
-      <form :disabled="pending" @submit.prevent="onSubmit">
+      <form 
+          ref="form"
+          :disabled="pending" @submit.prevent="onSubmit" 
+          @keydown.caps-lock="capsLock"
```
Mogge, line 17: When playing around I felt that
```
@keydown="capsLock"
@keyup="capsLock"
```
Has a much better behavior because we check on every key event.

webapp/components/LoginForm/LoginForm.spec.js
```diff
@@ -57,5 +58,20 @@ describe('LoginForm', () => {
...
+      })
+
+      it('shows warning when caps-lock is press', async () => {
```
Mogge, line 69: `is pressed`?

Mogge (changes requested):
This is really a difficult task. I have been playing around for two hours, and I cannot find any better solution. It is working, a bit buggy, but it works.

webapp/components/LoginForm/LoginForm.vue
Brandon-G-Tripp: @Mogge Is there an abbreviation in german for the capslock? Or should we use a symbol instead? The localization is causing issues with the styling and its consistency between the locales.

Mogge: `Capslock ist an` could be an alternative

nimitbhargava (changes requested):
Tried the fix on my local. 
Have the below feedbacks - 

Issue 1: [If caps already on, then alert not shown](https://s8.gifyu.com/images/ocelot-pr-review-issue-1.gif) (I understand that it's already been mentioned in the PR; but I feel this being the most of the cases)
Issue 2: [When password input is not in focus, then caps lock changes are not tracked](https://s8.gifyu.com/images/ocelot-pr-review-issue-2.gif)
Issue 3: [The expected UI seems to be different in GitHub Issue UI](https://github.com/Ocelot-Social-Community/Ocelot-Social/issues/3779)

Mogge:
After a discussion on the issue, we decided to put it in the icebox. With the show password feature we have a tool that helps the user with any password problems.  The solution for the caps warning is so buggy, that it will generate a lot of bug reports. So it is better not to merge it.

PR #4083, opened 2020-12-17
feat: Node Version Recommendation In backend/README.md

Mogge:
## 🍰 Pullrequest
Adding a information for the node version in use by us and how to switch between different node versions locally.


Tirokk (approved):
Hey @Mogge !
This is great !!! 🚀💫

Thanks for investigating. 🙏🏼👍🏼🍀

Saving a post with an image is working with node 12.9.0 now.
🐛 I have only a little wrong spelling, but I do approve anyways.

PS: I’m pretty sure the clipping worked with PNG files as well, because I have done it several times, because my only test picts are two PNGs. May they where converted to a JPEG in the past somehow somewhen?

PR #4084, opened 2020-12-17
feat: 🍰 Create Default Admin User  

### LINKED ISSUE #4012: 💥 [DevOps] Create Network With Only An Adminuser
## :fire: DevOps ticket

Create Network with only an Adminuser
(currently its either empty or has seed data)
(Security analysis pls)

Mogge:
## 🍰 Pullrequest
When running `yarn db:migrate init` it checks if there is any user in the database. If there is none, a default admin user is created.

### Issues
- fixes #4012

### Todo

- [X] Ensure that the default admin can login


backend/src/db/migrate/store.js
```diff
@@ -1,11 +1,62 @@
...
+  password: hashSync('1234', 10),
+  name: 'admin',
+  id: uuid(),
+  slug: 'admin',
+}
+
+const createDefaultAdminUser = async (session) => {
+  const readTxResultPromise = session.readTransaction(async (txc) => {
+    const result = await txc.run(
+      'MATCH (email:EmailAddress)-[:BELONGS_TO]->(user:User) RETURN count(user) AS userCount',
```
ulfgebhardt, line 16: Why do you use the EMail relation here?
is it not sufficient to just count the users?

backend/src/db/migrate/store.js
```diff
@@ -1,11 +1,62 @@
...
+         })-[:BELONGS_TO]->(u:User {
+           name: "${defaultAdmin.name}",
+           encryptedPassword: "${defaultAdmin.password}",
+           role: "admin",
+           id: "${defaultAdmin.id}",
+           slug: "${defaultAdmin.slug}",
+           createdAt: toString(datetime()),
+           deleted: false,
+           disabled: false
+         })-[:PRIMARY_EMAIL]->(e)`,
```
ulfgebhardt, line 42: idealy we would make a file out of all queries instead of putting them in the code - this would have the potential benefit of reuse. Ignore this for now

ulfgebhardt (approved):

Tirokk (commented):
Heya! Cool stuff @Mogge 

That one we really need to get to the next step! 🚀🚀💫💫💫
To have `wir.social` as real network with our friends …

Well done !!! 😘
Sorry that I can not approve now, because I couldn’t get the migrations running on Kubernetes and I had some longer problem with Neo4j in the DO Kloster.

I have some questions and annotations …

## Questions

The categories are not created. Is this on purpose, because I have understand that you will create them? Are they created somewhere else in the migration?
!!! –>  We talked about this in our Daily Standup (23. Dec. 2020) and decided to create an additional PR to deactivate the categories on the post and see if all is running well.

In file `backend/README.md` there are the commands:
```bash
yarn run db:migrate init
docker-compose exec backend yarn run db:migrate init
```
If I run them I get:
```bash
  init : undefined
```
… more details:
```bash
$ yarn run __migrate --store ./src/db/migrate/store.js init
$ migrate --compiler 'js:@babel/register' --migrations-dir ./src/db/migrations --store ./src/db/migrate/store.js init
  migrations dir : /develop-backend/src/db/migrations
Successfully created database indices and constraints!
  init : undefined
```
If I leave `init` away then the command does not function at all …

## Annotations

 It looks like that in Travis the `db:init` comes before the `db:seed` without prior call of `db:reset`. This can cause a problem in case Travis will run once again.

Mogge:
I think the error messages shows up because `next` in the call of `init` is undefined. We can live with this messages if everything else is executed fine (which is the case as far as I can see)

Tirokk:
Because we are starting with categories, like we said in our Strategy Meeting today, shouldn’t we set the categories here now for the start as well @Mogge @ulfgebhardt ?

Tirokk (approved):
Hey @Mogge !

This is a big step to our `wir.social` .
Cool 🚀💫💫

PR #4086, opened 2020-12-18
Throw Error When No Cookie Can Be Set

### LINKED ISSUE #4080: 🐛 [Bug] No cookies - no errors
## :bug: Bugreport

If you do not accept any cookies you can login and get a success message, but you will not be able to go to the dashboard.
-> Mitigate by displaying an error message

Mogge:
## 🍰 Pullrequest
Checks if a cookie with the JWT exists after login. If not, an error message is toasted, saying that no cookie could be set.

### Issues
- fixes #4080


ulfgebhardt (approved):
Awesome! <3
works well, I beleive tests are missing - but i figure we can do that with coverage in the future

PR #4121, opened 2021-01-09
start_on_register

### LINKED ISSUE #4089: 🚀 [Feature] Page starts with Register rather then Login
## :rocket: Feature

When opening `wir.social` the Page starts with Register rather then Login.
This is since the User will stay logged in for a long time and is directly redirected to the Dashboard, but a new User, not having a Cookie will be presented immediately with the register page

This task consists of those subtasks
- make the default page start at `wir.social/register` instead of `wir.social/login`
- Prolong the validity of the cookie set on the device to 2 years (so we don't need 2 login every day)

This is sort of analog to facebook.

image
image

ulfgebhardt:
## 🍰 Pullrequest

- on logout force to go to login page
- start with register page

### Issues

- fixes #4089 

### Todo

- [x] Make followup ticket for proper paths `registration/signup` must become `register` or `signup`
- [x] Make followup ticket for refactoring pages folder - it contains components and is not clean
- [x] Make followup ticket to prolong Cookie time to 2 years (@Mogge has done that?)
- [x] Make followup ticket for making start page & cookie time configurable
- [x] Make followup ticket for refactoring config

### DIFF
--- a/webapp/middleware/authenticated.js
+++ b/webapp/middleware/authenticated.js
@@ -22,5 +22,5 @@ export default async ({ store, env, route, redirect }) => {
     params.path = route.path
   }
 
-  return redirect('/login', params)
+  return redirect('/registration/signup', params)
 }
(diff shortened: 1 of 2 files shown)


Tirokk (approved):
Great @ulfgebhardt !!!
We get even more close to open up `wir.social`. 🚀💫

To do all your yet unchecked tickets seems the right way to me! 🍀

PR #4125, opened 2021-01-11
feat: Invite codes

Mogge:
## 🍰 Pullrequest
Backend part for using invite codes for user registration


ulfgebhardt:
As far as I can see our change we did together is missing?! @Mogge

Mogge:
What changes do you mean @ulfgebhardt ?

backend/src/schema/resolvers/helpers/generateInviteCode.js
```diff
@@ -0,0 +1,5 @@
+export default function generateInviteCode() {
+  return Array.from({ length: 6 }, (n = Math.floor(Math.random() * 36)) => {
+    return String.fromCharCode(n > 9 ? n + 55 : n + 48)
```
ulfgebhardt, line 3: Here a comment would be awesome ;). Took me a sec to understand.
Also `55` is incorrect - should be `65`. See https://en.wikipedia.org/wiki/ASCII

```suggestion
    // Shift ASCII Characters to numbers (48-57) or A-Z (65-90)
    return String.fromCharCode(n > 9 ? n + 65 : n + 48)
```

backend/src/schema/resolvers/inviteCodes.js
```diff
@@ -0,0 +1,109 @@
...
+        session.close()
+      }
+    },
+    isValidInviteCode: async (_parent, args, context, _resolveInfo) => {
+      const { code } = args
+      if (!code) return false
+      const session = context.driver.session()
+      const readTxResultPromise = session.readTransaction(async (txc) => {
+        const result = await txc.run(
+          `MATCH (ic:InviteCode { code: $code })
```
ulfgebhardt, line 43: The user input should be caps-agnostic - so Something like this should do the trick (Pseudocode, not tested)

```suggestion
          `MATCH (ic:InviteCode { code: UPPER($code) })
```

backend/src/schema/resolvers/inviteCodes.spec.js
```diff
@@ -0,0 +1,191 @@
...
+      })
+      const code = inviteCode.get('code')
+      const result = await query({ query: isValidInviteCodeQuery, variables: { code } })
+      expect(result.data.isValidInviteCode).toBeFalsy()
+    })
+
+    it('does not validate an invite code which does not exits', async () => {
+      const result = await query({ query: isValidInviteCodeQuery, variables: { code: 'AAA' } })
+      expect(result.data.isValidInviteCode).toBeFalsy()
+    })
```
ulfgebhardt, line 189:
```suggestion
    })
    
    it('does validate an invite code in lower caps' , async () => {
    ...
    }
```

ulfgebhardt (changes requested):
Awesome work! Just a few pit falls.
- [ ] Ascii encoding incorrect?
- [ ] Lower caps Codes acceptance?

backend/src/schema/resolvers/helpers/generateInviteCode.js
Mogge: No. It is `return String.fromCharCode(n > 9 ? n + 55 : n + 48)` because 10 + 55 = 65

Mogge: But I agree in commenting this.

backend/src/schema/resolvers/inviteCodes.js
Mogge: Great. Thank you!

backend/src/schema/resolvers/inviteCodes.spec.js
Mogge: Yep. This one is missing.

Mogge: See line 166

Tirokk (approved):
Yeah @Mogge 
This is an important peace of work to reach our GO LIVE! 🚀🚀💫

Great and sustainable work you have done.
I suggested just some easy changes and asked some not as important questions …

Wisdom

PR #4148, opened 2021-01-19
feat: Auto Detect Language Of Posts

### LINKED ISSUE #4147: 🚀 [Feature] Auto detect languages
## :rocket: Feature
To improve user experience we should get rid of the language field in the contribution form. On the other hand and would be nice to have the language of a post. So we should detect it automatically.

Mogge:
## 🍰 Pullrequest
Using detect language package (https://www.npmjs.com/package/languagedetect) to detect the language of a post automatically.

### Issues
- fixes #4147

### DIFF
--- a/backend/src/middleware/languages/languages.js
+++ b/backend/src/middleware/languages/languages.js
@@ -1,18 +1,15 @@
 import LanguageDetect from 'languagedetect'
 
 const setPostLanguage = (text) => {
-  console.log(text)
   const lngDetector = new LanguageDetect()
   lngDetector.setLanguageType('iso2')
   const result = lngDetector.detect(text, 2)
-  console.log(result)
   return result[0][0]
 }
 
 export default {
   Mutation: {
     CreatePost: async (resolve, root, args, context, info) => {
-      console.log('CreatePost, language', args)
       args.language = await setPostLanguage(args.content)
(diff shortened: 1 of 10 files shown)


backend/src/middleware/languages/languages.js
```diff
@@ -0,0 +1,42 @@
...
+  return sanitizeHtml(input, {
+    allowedTags: [],
+    allowedAttributes: {},
+  })
+}
+
+const setPostLanguage = (text) => {
+  const lngDetector = new LanguageDetect()
+  lngDetector.setLanguageType('iso2')
+  const result = lngDetector.detect(removeHtmlTags(text), 2)
```
ulfgebhardt, line 14: Will this always return two results? Because else the lines below could fail, since `result` might not contain a first and second element.

ulfgebhardt (commented):
I think it would be wise to have the language detection in the client. The detection could pre-select  a language and leave the final decision to the user. Assuming the language detection would produce `english` instead of `walish` could be corrected by the user.

What you think @Mogge ?

ulfgebhardt:
- [ ] Remove everything but the first language detected (no scores, no second langauge)
- [ ] Consider to save the language instead of country code

ulfgebhardt (approved):

Issue #4092, opened 2020-12-21
🚀 [Feature] Redesign Registration Process


ulfgebhardt:
## :rocket: Feature

It might be useful to implement the backend frist and then go for the frontend.
Features required

- EMail Confirmation, Change/Adjustment of the existing system
- Invite Code (generation, distribution, accepting)
- Invite Code Database Structure
- User interface like shown below

### User Interface

1. Invite Code

![register1 desktop_1920x1080](https://user-images.githubusercontent.com/1238238/102780200-d05ee480-4395-11eb-92f5-29bbd857100a.png)

2. EMail

![register2 desktop_1920x1080](https://user-images.githubusercontent.com/1238238/102780197-cfc64e00-4395-11eb-8d67-835432916f9d.png) 

3. EMail Code

![register3 desktop_1920x1080](https://user-images.githubusercontent.com/1238238/102780195-cf2db780-4395-11eb-862a-747306cb2a91.png)

4. Username

![register4 desktop_1920x1080](https://user-images.githubusercontent.com/1238238/102780193-ce952100-4395-11eb-9b03-2b021eb0efb0.png)

5. Checks

![register5 desktop_1920x1080](https://user-images.githubusercontent.com/1238238/102780190-cdfc8a80-4395-11eb-9788-fa73dd7f65fd.png)

6. Password

![register6 desktop_1920x1080](https://user-images.githubusercontent.com/1238238/102780188-cc32c700-4395-11eb-8a7b-6781418b92e3.png)

### Program Structure

![register flow](https://user-images.githubusercontent.com/1238238/102780411-2895e680-4396-11eb-9286-c953351e46fa.png)

### Notes

This Issue is very big - feel free to make subissues or pullrequests fixing some of this

PR #4168, opened 2021-01-27
feat: 🍰 Redesign Registration Process Frontend

Tirokk:
## 🍰 Pullrequest

Redesign the frontend of the registration process.

### Issues

- relates #4092

### Todo

- [ ] a lot


Tirokk:
I added more functionality. @Mogge

webapp/components/Registration/RegistrationItemEnterNonce.vue
```diff
@@ -0,0 +1,107 @@
...
+        },
+      })
+    },
+    async handleInput() {
+      this.sendValidation()
+    },
+    async handleInputValid() {
+      this.sendValidation()
+    },
+    handleSubmitVerify() {},
```
Tirokk, line 93: A database request would be nice here as for the invite.
Or what you think. @Mogge @ulfgebhardt

Tirokk (commented):
Please do a pre-review. @Mogge @ulfgebhardt 

This is my own pre-review to give you some hints about questions and leftovers and have some reminders for me.

- I haven't done translations in case we will have general changes in the texts …
- Comments marked with ‚Wolle‘ have do be proven and probably deleted.
But this is not sure at the moment.

webapp/components/ComponentSlider/ComponentSlider.vue
```diff
@@ -0,0 +1,87 @@
...
+    async onNextClick() {
+      let success = true
+      if (this.sliderData.sliders[this.sliderIndex].button.sliderCallback) {
+        success = await this.sliderData.sliders[this.sliderIndex].button.sliderCallback()
+      }
+      success = success && this.sliderData.sliders[this.sliderIndex].button.callback(success)
+      if (success && this.sliderIndex < this.sliderData.sliders.length - 1) {
+        this.sliderData.sliderSelectorCallback(this.sliderIndex + 1)
+      }
+    },
```
ulfgebhardt, line 71: This whole function seems a bit omplicated. Why three functions `sliderCallback`, `callback` and `sliderSelectorCallback` ?

webapp/components/Registration/RegistrationItemCreateUserAccount.vue
```diff
@@ -0,0 +1,321 @@
+<template>
```
ulfgebhardt, line 1: I really would love to see this page split into several pages as suggested by the design. What do you think?

-> Furthmore I believe that those componenents should be named something like `RegisterationSlide` or `RegistrationPage` not `RegistrationItem`

webapp/components/Registration/RegistrationSlider.vue
```diff
@@ -0,0 +1,235 @@
...
+      </template>
+
+      <component-slider :sliderData="sliderData">
+        <template #header>
+          <ds-heading size="h2">
+            {{ $t('components.registration.signup.title', metadata) }}
+          </ds-heading>
+        </template>
+
+        <template v-if="['invite-code'].includes(registrationType)" #enter-invite>
```
ulfgebhardt, line 17: shouldn't it be `registrationType.includes('invite-code')` ?

webapp/components/Registration/RegistrationSlider.vue
```diff
@@ -0,0 +1,235 @@
...
+          nonce: null,
+          name: null,
+          password: null,
+          passwordConfirmation: null,
+          about: null,
+          termsAndConditionsConfirmed: null,
+          dataPrivacy: null,
+          minimumAge: null,
+          noCommercial: null,
+          noPolitical: null,
```
ulfgebhardt, line 167: maybe we can put that into each slider and just have a array of sliderdata objects. This is not needed for this PR tho - just as reminder for refactoring

ulfgebhardt (commented):
Great work - code looks pretty clean. Please see my comments - but I believe we can push that through next week <3

webapp/components/ComponentSlider/ComponentSlider.vue
Tirokk: We invent a more general slider component and therefore I guess it is not always clear if decisions about what will happen on a button click can always only be done in the deepest level.
Therefore I implement a mechanism to have the possibility to decide at top-level as well or in combination. @ulfgebhardt

webapp/components/Registration/RegistrationItemCreateUserAccount.vue
Tirokk: I have not to much heart in that it must be this big. And the base card could have a bit more width to give the user a better overview.
This way it is split into kind of logical steps …

I would like to rename the components as well. 😍👍🏼

webapp/components/Registration/RegistrationSlider.vue
Tirokk: I’ve used the same style as for the other slider components to have a better recognition that’s about the same thing. Could be for sure as well:
```suggestion
        <template v-if="registrationType === 'invite-code'" #enter-invite>
```
 As far I've understand the array comes before the dot selector and `registrationType` is not an array …

webapp/components/Registration/RegistrationSlider.vue
Tirokk: I think it's not needed for code functionality anyways. For me it has the functionality to give an overview about all the data to collect at the top-level.

Mogge (approved):

PR #4185, opened 2021-01-29
Fix locations

ulfgebhardt:
### Issues

- fixes https://github.com/Ocelot-Social-Community/Ocelot-Social/issues/4183

### Todo

- [x] Tests - I want a lesson if you don't mind @Mogge 
- [x] Issue for Mixin (do we really need that?)


backend/src/schema/resolvers/users/location.spec.js
```diff
@@ -85,6 +95,62 @@ beforeEach(() => {
 
 afterEach(cleanDatabase)
 
+describe('Location Service',() => {
+  // Authentication
+  // TODO: unify, externalize, simplify, wtf?
+  let user
+  beforeEach(async () => {
```
Mogge, line 102: Why beforeEach? We can use the same user for all following it-blocks. So I would suggest beforeAll to create the user. Then may be beforeEach to set the authenticated user (it is set to null beforeEach in the head), or set the authenticated user in each async it-block explecitely

backend/src/schema/resolvers/users/location.spec.js
```diff
@@ -85,6 +95,62 @@ beforeEach(() => {
...
+    user = await Factory.build('user', {
+      id: 'updating-user',
+    })
+    authenticatedUser = await user.toJson()
+  })
+  it('query Location existing', async () => {
```
Mogge, line 108: Cosmetics: Please one blank line before each new it-bock

Mogge (commented):
I really like this fix! I always thought that doing the axios call in the frontend is no good idea. I am happy we run into this.
I have some cosmetic suggestions which you can discard easily. I would ask you to create another issue, addressing the location query in webapp, to have it as a plugin/mixin. I am sure we need the same code in other occasions again.

ulfgebhardt:
@Mogge I wrote a lot of frontend tests for this - please review!

ulfgebhardt:
@Mogge Your suggestion to use `beforeAll` instead of `beforeEach` seemed reasonable and logical, but doesn't work.
Since I am very unhappy with how user authentication is mocked anyways, I recommend to keep using `beforeEach` for now, since that seems to be working.

backend/src/schema/types/type/Location.gql
```diff
@@ -16,3 +16,12 @@ type Location {
   parent: Location @cypher(statement: "MATCH (this)-[:IS_IN]->(l:Location) RETURN l")
 }
 
+type Location2 {
```
Mogge, line 19: We need another name: `LocationHelper` or `LocationFinder`

backend/src/schema/types/type/Location.gql
```diff
@@ -16,3 +16,12 @@ type Location {
...
+  id: ID!
+  place_name: String!
+}
+
+type Query {
+  queryLocations(place: String!, lang: String!): [Location2]!
```
Mogge, line 25: Change the name of the result.

webapp/pages/settings/index.spec.js
```diff
@@ -96,6 +127,116 @@ describe('index.vue', () => {
           expect(mocks.$apollo.mutate).toHaveBeenCalled()
```
Mogge, line 127:
```
expect(mocks.$apollo.mutate).toHaveBeenCalledWith(
            expect.objectContaining(
              {
                variables: expect.objectContaining(
                  {
                    name: 'Peter', 
                  },
                ),
              }
            )
          )
```

webapp/pages/settings/index.spec.js
```diff
@@ -96,6 +127,116 @@ describe('index.vue', () => {
...
       })
+
+      describe('given a new slug and hitting submit', () => {
+        it('calls updateUser mutation', () => {
+          const wrapper = Wrapper()
+
+          wrapper.find('#slug').setValue('peter-der-lustige')
+          wrapper.find('.ds-form').trigger('submit')
+
+          expect(mocks.$apollo.mutate).toHaveBeenCalled()
```
Mogge, line 138:
```javascript
expect(mocks.$apollo.mutate).toHaveBeenCalledWith(
            expect.objectContaining(
              {
                variables: expect.objectContaining(
                  {
                    slug: 'peter-der-lustige', 
                  },
                ),
              }
            )
          )
```

webapp/pages/settings/index.spec.js
```diff
@@ -96,6 +127,116 @@ describe('index.vue', () => {
...
+      })
+
+      describe('given a new location and hitting submit', () => {
+        it('calls updateUser mutation', () => {
+          const wrapper = Wrapper()
+
+          wrapper.find('#city').setValue('Berlin, Germany')
+          wrapper.find('.ds-form').trigger('submit')
+
+          expect(mocks.$apollo.mutate).toHaveBeenCalled()
```
Mogge, line 149: Same here

webapp/pages/settings/index.spec.js
```diff
@@ -96,6 +127,116 @@ describe('index.vue', () => {
...
+      })
+
+      describe('given a new about and hitting submit', () => {
+        it('calls updateUser mutation', () => {
+          const wrapper = Wrapper()
+
+          wrapper.find('#about').setValue('I am Peter!111elf')
+          wrapper.find('.ds-form').trigger('submit')
+
+          expect(mocks.$apollo.mutate).toHaveBeenCalled()
```
Mogge, line 160: same here

webapp/pages/settings/index.spec.js
```diff
@@ -96,6 +127,116 @@ describe('index.vue', () => {
...
+        it('calls updateUser mutation', () => {
+          const wrapper = Wrapper()
+
+          wrapper.find('#name').setValue('Peter')
+          wrapper.find('#slug').setValue('peter-der-lustige')
+          wrapper.find('#city').setValue('Berlin, Germany')
+          wrapper.find('#about').setValue('I am Peter!111elf')
+          wrapper.find('.ds-form').trigger('submit')
+
+          expect(mocks.$apollo.mutate).toHaveBeenCalled()
```
Mogge, line 174: same here

webapp/pages/settings/index.spec.js
```diff
@@ -96,6 +127,116 @@ describe('index.vue', () => {
...
+        const wrapper = Wrapper()
+
+        jest.useFakeTimers()
+
+        wrapper.find('#city').trigger('input')
+        wrapper.find('#city').setValue('B')
+
+        jest.runAllTimers()
+
+        expect(mocks.$apollo.query).toHaveBeenCalled()
```
Mogge, line 190: Same here. Did you set the locale?

webapp/pages/settings/index.vue
```diff
@@ -123,51 +121,38 @@ export default {
...
-    requestGeoData(e) {
-      if (this.axiosSource) {
-        // cancel last request
-        this.axiosSource.cancel()
-      }
-
+    async requestGeoData(e) {
       const value = e.target ? e.target.value.trim() : ''
-      if (value === '' || value.length < 3) {
+      if (value === '') {
```
Mogge, line 141: Are you sure you want to start the query with just one char of input? I would prefer starting the query if we have at least 3 chars of input.

webapp/pages/settings/index.spec.js
```diff
@@ -27,6 +28,36 @@ describe('index.vue', () => {
               },
             },
           }),
+        query: jest
+          .fn()
+          .mockRejectedValue({ message: 'Ouch!' })
+          .mockResolvedValueOnce({
+            data: {
+              queryLocations: [
+                { place_name: 'Brazil', id: 'country.9531777110682710', __typename: 'Location2' },
```
Mogge, line 37: Change the typename.

webapp/pages/settings/index.spec.js
```diff
@@ -27,6 +28,36 @@ describe('index.vue', () => {
...
+                {
+                  place_name: 'United Kingdom',
+                  id: 'country.12405201072814600',
+                  __typename: 'Location2',
```
Mogge, line 41: change

webapp/pages/settings/index.spec.js
```diff
@@ -27,6 +28,36 @@ describe('index.vue', () => {
...
+                },
+                {
+                  place_name: 'Buenos Aires, Argentina',
+                  id: 'place.7159025980072860',
+                  __typename: 'Location2',
```
Mogge, line 46: etc.

backend/src/schema/types/type/Location.gql
Mogge: In some months we will not remember what Location2 means

Mogge (changes requested):
Please find a better name for Location2. Something more meaningful. I gave some examples how to test the variables passed to the mutation.

ulfgebhardt:
> Please find a better name for Location2. Something more meaningful. I gave some examples how to test the variables passed to the mutation.

I did this because there is no option to comment in gql files - it makes zero sense to have two different location objects - especially in a graph database. So I am glad you noticed - I'd like to change it and make one out of it - but I believe its a bit out of scope for this PR. Would be an Issue for it acceptable?

backend/src/schema/types/type/Location.gql
ulfgebhardt: > Please find a better name for Location2. Something more meaningful. I gave some examples how to test the variables passed to the mutation.

I did this because there is no option to comment in gql files - it makes zero sense to have two different location objects - especially in a graph database. So I am glad you noticed - I'd like to change it and make one out of it - but I believe its a bit out of scope for this PR. Would be an Issue for it acceptable?

ulfgebhardt: As I understand it currently the user object just holds the name of the place he has set in his language - thats a nice touch - so if you set "München" everyone in the world will see "München" instead of Munich - but if we at any time want to display the location on a map that is more then impractical

webapp/pages/settings/index.spec.js
ulfgebhardt: Locale is set statically to en inf queried within vue - see mocks

webapp/pages/settings/index.vue
ulfgebhardt: For what reason?

backend/src/schema/types/type/Location.gql
ulfgebhardt: using `LocationMapBox`

ulfgebhardt:
Two test fail - the city tests since the select text change does not trigger a model change - its not in the select options since they are queried from the server - how to mock this @Mogge ?

Mogge (approved):
I fixed the test for location changes. Feel free to merge. Thank you very much for this PR! It is a great improvement.

Mogge:
@ulfgebhardt I set the cities data of the component and simulated a click on the dropdown  menu to set the new location value.

ulfgebhardt:
> @ulfgebhardt I set the cities data of the component and simulated a click on the dropdown menu to set the new location value.

I see now - you operate the select - sorry to have burdened you with this - learned something here <3

PR #4194, opened 2021-02-03
Clean GitHub folder

ulfgebhardt:
## 🍰 Pullrequest

Just some cleanups of the `.github` folder in preparation for implementing a publish workflow

### Issues

- None

### Todo

- [X] None


.github/semantic.yml
```diff
@@ -1,2 +0,0 @@
-# Always validate the PR title, and ignore the commits
-titleOnly: true
```
Tirokk, line 2: Why you remove this? @ulfgebhardt 
If you remove it then the semantic-PR fails if you have just one commit. One needs then at least two commits what is mind stretching.

I suggest to leave it in …

Tirokk (commented):
Hey! Nice changes! @ulfgebhardt 
I appreciate you caring !!! 🚀🚀💫🌟

I would suggest to leave the semantic PR YAML in. See comment …

Everything else looks good to me!

.github/semantic.yml
Tirokk: Or have you switch the semantic-PR off?
If so, please tell me why you don’t want them. They give a better order in the change log as far I remember …

Mogge (approved):
And yet another step forward. Thank you so much.

.github/semantic.yml
ulfgebhardt: I do not understand sematic pr, can you elabirate on that? I have not switched anything off.

ulfgebhardt: This seems to be an addon, not default github standard.

There is two approaches:
- github actions: https://github.com/marketplace/actions/semantic-pull-request
- github app: https://github.com/apps/semantic-pull-requests

I think this might be nice for the changelog, but its definitely not working at the moment.
@Tirokk Please make an Issue to enable it - then we should do a reconfigure anyways.

PR #4195, opened 2021-02-03
publish workflow

ulfgebhardt:
## 🍰 Pullrequest

- publish to dockerhub
- publish to github release
- release command in main package.json to bump all versions & generate changelog `yarn release`
- build tags on github & dockerhub for unique build images e.g. `0.6.5.22` where the 22 refers to the CI Build Number. It is persistent across versions

Problem which will not be solved now: The github release cannot be updated  while the dockerhub release will. Therefore the github release will contain the first version and the dockerhub the last version for one version-iteration. Not very intuitive, but I guess that's how it goes. It is recommended to use specific build versions when deploying.

Problem which will not be solved now: Releases appear as unverified

The last commit will always be omitted and therefore should always just include the version bump

This PR includes two version bumps 0.6.4 & 0.6.5 -> we cannot remove them from the releases I guess ;) lets keep them

image

image

image

image

image


Mogge (approved):
Great! Thank you

PR #4215, opened 2021-02-11
Move Old Migrations From HC To Examples Folder To Avoid Conflicts

Mogge:
## 🍰 Pullrequest
Move the old migrations in the examples folder to avoid conflicts with new migrations.


Tirokk (approved):
Really great you thought about that! @Mogge 😍

Relief

PR #4220, opened 2021-02-15
fix: Location Tests In Backend

Mogge:
MAPBOX introduced district for US-Cities. So I used a German City instead


Tirokk (approved):
Thanks for your quick response @Mogge !!!
YEAH !!! 🚀💫🍀

PR #4225, opened 2021-02-16
fix: Query Available Roles As Admin

Mogge:
## 🍰 Pullrequest
avoid introspection to get available roles as admin


ulfgebhardt (approved):

ulfgebhardt:
consider to make a new version for this, so we can deploy it `yarn release`

PR #4230, opened 2021-02-17
fix: css so follow button isn't blocked by adblock

Mogge:
## 🍰 Pullrequest
css fix to trick adblock on follow button by @nila99


ulfgebhardt (approved):

Tirokk (approved):
Let’m go ! @Mogge

PR #4332, opened 2021-04-01
fix: [WIP] 🍰 Follow Relation Not Set At Invite

Tirokk:
## 🍰 Pullrequest

Fix follow relation not set by invite in mutation `SignupVerification`.

### Issues

- fixes #4325

### Todo

- [x] find out the problem

PR #4336, opened 2021-04-02
fix: Email Confirmation-Link When An Invite-Code Is Given

### LINKED ISSUE #4325: 🐛 [Bug] Follow Relation Is Not Set At Invite
## :bug: Bugreport

Follow relation is not set at invite.

- In the dev environment the relation is set, but the `INVITED` and the `FOLLOWS` relations have not `createdAt` property.
- Live there is no `INVITED` or `FOLLOWS` relation at all.

### Closer Problem

If the user in the registration process gets an e-mail verification e-mail and clicks the link then there is not invite code in it.
This causes the problem. In dev I get no real e-mail and therefore I don’t click a link and stay in the same registration slider configuration where the invite code is typed in.

### Todo

- [ ] set the relation to the invite node from the `emailAddress` node.
- [ ] care for `Wolle` comments

### LINKED ISSUE #4339: 🐛 [Bug] Email Verification Sends Out Wrong Link
## :bug: Bugreport

The email sent to verify the email address during signup contains a wrong link.
When starting here:

email_entered

You can see, that there is no method given in the URL.

This is en extract of the email the new user receives (please ignore the `\n +`, I have overseen this when cleaning up):

email

As you can see, there is a huge button shouting **CLICK ME** in your face, and further below, you see the actual nonce you are asked to enter during signup process.

There are very few users in this world, which would not click on the button and type in the nonce in the form instead.

But now, this is the link the button is pointing to:
```
http://localhost:3000/registration?method=invite-mail&email=test%40test.org&nonce=71967
```
And this  is what you see, when you click on it:

nonce

You see, now we have a method in the URL, and it is the wrong one! We switched over to another signup method.

The dots do go back to the earlier steps of the registration process disappeared!

The backend has no chance to receive the invite code from this point on, because it does not exist in the frontend anymore. 

### Steps to reproduce the behavior
1. Got to `http://localhost:3000/registration`
2. Add a line `console.log(actionUrl)` to line 24 of `backend/middleware/email/templateBuilder.js` to see the link in the console.
3. Add a line `console.log(signupTemplate({ email, nonce })).html` to line 48 of `backend/src/middleware/email/emailMiddleware.js` to see the email in the console.
4. Step forward in the registration process until you get the email send.
5. Check the Link and the Email

### Expected behavior
The Link should point to the correct signup method or no link should be provided at all.

Mogge:
## 🍰 Pullrequest
When an invite code is passed to the email verification, the invite code is added to the action URL and the method is changed to invite-code

### Issues
- fixes #4339
- fixes #4325
- relates #4332


ulfgebhardt:
I understand that this is optional, but didn't we agree that the invite code is not part of the EMail? @Mogge

Tirokk:
We have agreed that the invite code gets stored via `Signup` mutation as a relation from the `EmailAddress` node to the `InviteCode` node.
The `SignupVerification` mutation then uses this relation to find if the invite code was used for signup and generates the necessary relations. @Mogge 
… as far I know.

webapp/pages/registration.vue
```diff
@@ -39,18 +40,25 @@ export default {
...
+            activePage: null,
+          }
         )
       } else {
         if (
           this.method === 'invite-mail' ||
           (this.method === 'invite-code' && this.inviteRegistration)
         ) {
-          return this.method
+          if (this.method === 'invite-code' && this.inviteCode && this.nonce && this.email)
```
Tirokk: The problem here is that `this.inviteCode`, `this.nonce`, and `this.email` just do not exist, because they are only local variables in `data`.

We just have overlooked this. 😅

Tirokk: I’ve done an additional commit …

Mogge: But how did it work then when we tried it? @Tirokk

ulfgebhardt (approved):
Looks all good  to me, can we proceed?

Tirokk:
We can proceed. Shall I do a new Version or shall the version 1.0.2 be removed?
How we will handle this? @ulfgebhardt

PR #4980, opened 2022-05-15
fix: 🍰 Replace Hashtag In d.tube Url

### LINKED ISSUE #4896: 🐛 [Bug] Pasting Of Some Embedded Links Does Not Work
## :bug: Bug Report

- Pasting of the following d.tube link does not work correctly: `https://d.tube/#!/v/secretman02/d20zqahvv7d`

ogerly:
## 🍰 Pullrequest
we use the oembed library, in the provider.json ... you can see what is used. 
I cannot say at this point whether this is current. 
As far as d.tube is concerned, nothing has changed. 
but in the url you can copy from the browser is a /#!/ 
this causes the problem. 
the only solution i found now is to remove this. 

but i don't think this is a good solution. it should happen in the provider.json ... but there we have some problem in the code ... 

the hashtag function is not the problem

### Issues
- fixes #4896 

note: 
if a link from facebook shows the message 'please log in', then the user from whom this source comes has set it that way.

### DIFF
--- a/webapp/components/Editor/nodes/Hashtag.js
+++ b/webapp/components/Editor/nodes/Hashtag.js
@@ -8,7 +8,7 @@ export default class Hashtag extends TipTapMention {
   get defaultOptions() {
     return {
       matcher: {
-        char: '#',
+        char: '=',
         allowSpaces: false,
         startOfLine: false,
       },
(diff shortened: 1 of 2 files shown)


ogerly:
for facebook is a different problem ... if a user sets his account settings ... that you can only see his data internally ... then an included source is possible for everyone behind the login ... otherwise not.

Tirokk (approved):
Greate that you cared for this missfunction !!! 🚀💫

PR #5073, opened 2022-07-13
chore: 🍰 Change `image` Entries In Docker Compose Files And Fix Apple M1 Problem

### LINKED ISSUE #5072: 💥 [DevOps] Remove `image` Entries In Docker Compose Files And Get Apple M1 Running
## 💥 DevOps Ticket

Remove `image` entries in:

- `docker-compose.override.yml`
- `docker-compose.test.yml`?
- `docker-compose.yml`

Get Apple M1 running.

Because:

- in case there would be a Docker image in the repository it will be downloaded instead of building it fresh
- in newer Docker or Docker Desktop versions an error is shown if the image is not available in the image repo, see below

!!! add screen shot

Tirokk:
## 🍰 Pullrequest

Remove `image` entries in docker compose files and fix Apple M1 problem.

### Issues

- fixes #5072

### Questions

- [x] can the `image` entries really be removed from `docker-compose.yml`?
  - [x] or do we have to override them by emptying them somehow by `docker-compose.override.yml`?
    - [x] I just renamed them all
- [x] remove `image` entries from `docker-compose.test.yml` as well?
  - [x] no, they will stay
- [ ] may change command `docker-compose` to new `docker compose` notation?
- [x] fix Apple M1 Docker Compose build problem, see https://stackoverflow.com/questions/65612411/forcing-docker-to-use-linux-amd64-platform-by-default-on-macos
  - [x] may add `docker-compose.apple-m1.override.yml` for overriding?
    - [x] refine docs with Apple M1 informations and the new command?
      - `$ docker-compose -f docker-compose.yml -f docker-compose.override.yml -f docker-compose.apple-m1.override.yml up`
  - [x] after that `export DOCKER_DEFAULT_PLATFORM=linux/amd64` is an other possibility that work for me
    - [ ] should be set then in a global and stable manner

### Todo

- [x] change docker compose `maintenance` service port from `5000` to `3001`
  - because it is said that macOS has a standard use of  port`5000`


docker-compose.override.yml
```diff
@@ -5,7 +5,7 @@ services:
   # WEBAPP ###############################################
   ########################################################
   webapp:
-    image: ocelotsocialnetwork/webapp:development
+    # image: ocelotsocialnetwork/webapp:development
```
Tirokk, line 8: Remove the out commented images?

Tirokk: I renamed them all!

ogerly (approved):

PR #5102, opened 2022-07-25
feat: Make Categories Optional

Mogge:
## 🍰 Pullrequest
Include the old category mechanism, if the option `ACTIVE_CATEGORIES` is true.

### Issues

- None

### Todo

- [X] None


Mogge:
May be we should test at least the webapp with categories active.

ogerly (approved):

ogerly (approved):

Tirokk (approved):
Great to have categories back! 🚀💫🤗

Would be nice to have some tests with categories and to get rid of `Hc-*` …
But that's for the future –

PR #5198, opened 2022-08-22
feat: Categories Filter Menu

Mogge:
## 🍰 Pullrequest
Implement the category filter menu as it was in HC. Seed categories (when active) on `db:migrate init`.
Set categories as desired by Yunite

### Issues

- None

### Todo

- [X] None


backend/src/db/migrate/store.js
```diff
@@ -10,6 +12,28 @@ const defaultAdmin = {
...
+      const id = `cat${index + 1}`
+      txc.run(
+        `MERGE (c:Category {
+          icon: "${icon}",
+          slug: "${name}",
+          name: "${name}",
+          id: "${id}",
+          createdAt: toString(datetime())
+        })`,
+      )
```
Tirokk, line 27: What is if there are already categories?

Do we skip this merge?
Have a logging for it?

backend/src/db/migrate/store.js
```diff
@@ -45,7 +69,7 @@ const createDefaultAdminUser = async (session) => {
     })
     try {
       await createAdminTxResultPromise
-      console.log('Successfully created default admin user') // eslint-disable-line no-console
+      console.log('Successfully created default admin user!') // eslint-disable-line no-console
```
Tirokk, line 72: I allowed my self to put the exclamation mark behind the logging of created admin … as well and push …

backend/src/db/migrate/store.js
```diff
@@ -10,6 +12,28 @@ const defaultAdmin = {
...
+          slug: "${name}",
+          name: "${name}",
+          id: "${id}",
+          createdAt: toString(datetime())
+        })`,
+      )
+    })
+  })
+  try {
+    await createCategoriesTxResultPromise
```
Tirokk, line 31:
```suggestion
    await createCategoriesTxResultPromise
    console.log('Successfully created categories!') // eslint-disable-line no-console
```

Tirokk (changes requested):
May we can seed categories of some posts?
Otherwise we can not quickly check the filter …

We have to redesign the post teaser i guess, see screen shot. Or?

Bildschirmfoto 2022-08-25 um 18 33 33

May we should have the translations in the database?

Tirokk (approved):
Great @Mogge 🚀🚀💫💫

Issue #5131, opened 2022-08-01
🚀 [Feature] Implement Group GraphQL Model And CRUD Resolvers In Backend


Tirokk:
## :rocket: Feature Request

Implement group GraphQL model and CRUD resolvers.

- relates #5059

### Todo

- [x] GraphQL
  - [x] add group settings ( details see https://github.com/Ocelot-Social-Community/Ocelot-Social/issues/5059#user-content-group-settings )
    - [x] owner/admin
    - [x] name (title)
    - [x] slug-name
    - [x] about (goal)
    - [x] description:
      - [x] minimal 100 chars
    - [x] group type
    - [x] action radius
    - [x] location
    - [x] categories: 1-3
    - [x] add relation `CREATED`
  - [x] rename `UserGroup` to `UserRole`
  - [x] define resolver for:
    - [x] `CreateGroup`
      - [x] error on name length, see frontend
    - [x] `Group`
    - [x] `UpdateGroup`
    - [x] `DeleteGroup`, see #5388
- [x] seeding:
  - [x] seed a group

Tirokk:
solved

Issue #5059, opened 2022-07-04
🌟 [EPIC] Release v2.0.0 In Groups Branch - Group: For Groups/Organisations/Projects - Subset Of Former Orgs EPIC Named Group Now Instead Org


ogerly:
## 🌟 EPIC

——— This issue is is a subset of the EPIC #4544 ———

### General Concept

A group is the very general concept for every entity which is not a person – a group of people, an organisation, a business, a project, a charity, a foundation ...

Implement a feature called `group` means to create a new entity which is not a user account and therefore can not be logged in. But it has a group profile similar to a user profile, see below.

#### New `My Groups Page`

→ [more details](#user-content-my-groups-page)

Reach the `my groups` page:

- [x] New menu item `My groups`: ( #5395 )
  - [x] Chose this user menu of the header will open the page.

<img width="303" alt="Bildschirmfoto 2022-07-23 um 05 43 09" src="https://user-images.githubusercontent.com/25344101/180589335-7715d341-824b-4bca-be14-512da5d6e0fc.png">

##### Configurable Additional Options In Header

Option 1:

- [x] New icon with 2 people in a circle: (3 people would be more look like a group says @LarsE11 and @Tirokk agrees)
  - [x] Klick on this icon ([source stylguide](https://styleguide.human-connection.org/typography/dsicon)) in the header on the right hand between the language and the bell will open the page.

(see issue #5395 )

<img width="372" alt="Bildschirmfoto 2022-07-23 um 05 26 14" src="https://user-images.githubusercontent.com/25344101/180588868-585657dc-bb33-4b4c-9cd2-89b58f70b310.png">
<img width="415" alt="Bildschirmfoto 2022-07-23 um 05 30 59" src="https://user-images.githubusercontent.com/25344101/180589005-ad031f47-27c8-4628-bc55-de08bf2827b4.png">

Option 2:

( @LarsE11 and @UlfTramsen wish and it seem nice to have the possibility)

- [x] Configurable quick click menu between header logo on the right and the search field.
  - [ ] Maybe make the search field smaller:
    - [ ] May expand it if the cursor fokus is in it (or have just a lupe and make the search field animated expanding if you click it)
- [x] Have `Groups` or `My Groups` there as a text menu item.

[...]

PR #5224, opened 2022-08-30
feat: 🍰 Implement `UpdateGroup` Resolver

Tirokk:
## 🍰 Pull Request

Implement `UpdateGroup` resolver.

### Issues

- relates #5131
- relates #5059

### Todo

- [ ] GraphQL
  - [ ] add group settings ( details see https://github.com/Ocelot-Social-Community/Ocelot-Social/issues/5059#user-content-group-settings )
    - [x] categories
    - [x] avatar
    - [ ] location
- [ ] tests:
  - [x] general
  - [x] slugification
    - [x] if we change the name so that an existing slug will be generated that it is made unique by `*-1`
  - [x] test if resolver `Group` finds group by `id`
  - [ ] does avatar testing make sense this way in `backend/src/schema/resolvers/user_management.spec.js`
    - [ ] minimum: await is missing …
    - [ ] search for other appearances of `avatar: Factory.build('image',` or `Factory.build('image'` without await
    - [x] created an issue #5256 
- [x] cleanup: search for `Wolle`


backend/src/schema/resolvers/groups.js
```diff
@@ -131,6 +133,76 @@ export default {
         session.close()
       }
     },
+    UpdateGroup: async (_parent, params, context, _resolveInfo) => {
+      const { categoryIds } = params
+      const { id: groupId, avatar: avatarInput } = params
+      delete params.categoryIds
+      if (CONFIG.CATEGORIES_ACTIVE && categoryIds) {
```
Mogge, line 140: What happens when categories are active but `categoriesIds` is `undefined`

backend/src/schema/resolvers/groups.spec.js
```diff
@@ -2043,5 +2140,211 @@ describe('in mode', () => {
...
+              authenticatedUser = await user.toJson()
+            })
+
+            it('has set the settings', async () => {
+              await expect(
+                mutate({
+                  mutation: updateGroupMutation,
+                  variables: {
+                    id: 'my-group',
+                    name: 'The New Group For Our Coutry',
```
Mogge, line 2221: Mind the typo?

Mogge (changes requested):
Looks very good! Some question ...

backend/src/schema/resolvers/groups.js
Tirokk: If `categoriesIds` are `null` or `undefined` they are not changed in any way.

If `categoriesIds` are `[]` an error is thrown if `CONFIG.CATEGORIES_ACTIVE === true`.
If the array is not empty the old are deleted and new are set.

But there's no type check if it's an array. You thing it's needed?

Tirokk: The type check is done by GQL … 😉

backend/src/schema/resolvers/groups.spec.js
Tirokk: Nice catch! 🍀
done

Mogge (approved):

PR #5248, opened 2022-08-31
feat: Topics Menu

Mogge:
## 🍰 Pull Request
If categories are active, a menu point "Topics" appears to the left of the search field. Clicking on the menu, it opens a dropdown, where active topics can be selected


Tirokk (approved):
Jo cool @Mogge 😍

It works fine and I approve 🚀🚀💫

PR #5261, opened 2022-09-05
feat: Save Category Settings

Mogge:
## 🍰 Pull Request

Provide an API call, that allows an authenticated user to save the categories/topics with interest.

### DIFF
--- a/backend/src/schema/resolvers/users.spec.js
+++ b/backend/src/schema/resolvers/users.spec.js
@@ -3,6 +3,7 @@ import { gql } from '../../helpers/jest'
 import { getNeode, getDriver } from '../../db/neo4j'
 import createServer from '../../server'
 import { createTestClient } from 'apollo-server-testing'
+import { categories } from '../../constants/categories'
 
 const categoryIds = ['cat9']
 let user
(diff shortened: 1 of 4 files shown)


backend/src/schema/resolvers/users.spec.js
```diff
@@ -544,3 +551,98 @@ describe('switch user role', () => {
     })
   })
 })
+
+describe('save category settings', () => {
+  beforeAll(async () => {
```
Tirokk, line 556: May you mind to put this in the `beforeEach` here to get the skipped test running?

Because the db is cleaned `afterEach` test …

Works for me … 😁

Tirokk (changes requested):
Very cool!
I have some suggestions …

backend/src/schema/resolvers/users.spec.js
```diff
@@ -544,3 +551,98 @@ describe('switch user role', () => {
...
+          }),
+        )
+      })
+
+      describe('query for user', () => {
+        beforeEach(async () => {
+          await mutate({ mutation: saveCategorySettings, variables })
+        })
+
+        it.skip('returns the active categories when user is queried', async () => {
```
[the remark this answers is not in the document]
Mogge: The only works, but it does not work if all tests run :(

backend/src/schema/resolvers/users.spec.js
Mogge: Thanks for the hint

Tirokk (approved):
Yeah! 👍🏼

I approve 🚀🚀💫

PR #5284, opened 2022-09-06
feat: 🍰 Save Categories In Frontend

Mogge:
## 🍰 Pull Request
Provide a button in category filter menu, to save the current selected categories.


webapp/locales/de.json
```diff
@@ -363,7 +363,11 @@
         "label": "Älteste zuerst"
       }
     },
-    "order-by": "Sortieren nach ..."
+    "order-by": "Sortieren nach ...",
+    "save": {
+      "error": "Themen konnten nicht gespeichert werden",
```
Tirokk, line 368:
```suggestion
      "error": "Themen konnten nicht gespeichert werden!",
```

We put an exclamation mark behind error, I guess.

webapp/locales/en.json
```diff
@@ -363,7 +363,11 @@
         "label": "Oldest first"
       }
     },
-    "order-by": "Order by ..."
+    "order-by": "Order by ...",
+    "save": {
+      "error": "Failed saving topic settings",
```
Tirokk, line 368:
```suggestion
      "error": "Failed saving topic settings!",
```

We put an exclamation mark behind error, I guess.

Tirokk (changes requested):
In general it does not working as I would expect:

- After a reload of the page it loses all themes saved somehow.
- After login I don't have the themes set like I save for this user. Often I have the ones set which I had as I logged out with another user with different themes saved.

I have a little suggestion for the locales …

PS: In the database it is difficult to see which categories are selected on testing … 😅 Which is not related …

Tirokk (approved):
Very cool @Mogge 👍🏼😍

Now it works like expected !!! 🚀🚀💫💫

If you like create an issue for the reload problem and we have a look later on …

PR #5332, opened 2022-09-09
feat: 🍰 Implement Group Profile – Visibility

### LINKED ISSUE #5318: 🚀 [Feature] Implement Group Profile – Next Steps
## :rocket: Feature Request

Implement group profile – second step.

### Todo

- [x] member list
- [x] member count
- [x] join group button
- [x] leave group button
- [x] group type
- [x] action radius
- [x] my role
- [x] description
- [x] group location
- [ ] content menu
  - [x] show profile, settings ( #5419 )
  - [ ] report group ( #5449 )
- [x] categories
- [ ] avatar
  - [ ] is updating after upload? followup issue #5510
- [ ] invite button – followup issue #5511
- [ ] social-media
  - [x] created issue, see #5409 
- [x] tabs – removed
- [ ] test
  - [ ] backend
    - [ ] can only the users thierself join a group like via leave?
- [ ] does image upload on post still work?
- [x] `hidden` groups can only be seen by members that are not `pending`
- [ ] cleanup:
  - [ ] view comments in `webapp/pages/group/_id/_slug.vue`
  - [ ] search `Wolle`

Tirokk:
## 🍰 Pull Request

Implement group profile – second step.

### Issues

- relates #5318
- relates #5059

### Todo

- [x] `hidden` groups can only be seen by members that are not `pending`
- [x] cleanup:
  - [x] search `Wolle`


ogerly (approved):

backend/src/schema/resolvers/groups.js
```diff
@@ -10,30 +10,42 @@ import { mergeImage } from './images/images'
...
       const session = context.driver.session()
       const readTxResultPromise = session.readTransaction(async (txc) => {
-        const groupIdCypher = groupId ? ` {id: "${groupId}"}` : ''
+        const matchParamsEntries = Object.entries(matchParams)
+        let groupMatchParamsCypher = ''
+        matchParamsEntries.forEach((ele, index) => {
+          groupMatchParamsCypher += index === 0 ? ' {' : ''
+          groupMatchParamsCypher += `${ele[0]}: "${ele[1]}"`
+          groupMatchParamsCypher += index < matchParamsEntries.length - 1 ? ', ' : '}'
+        })
```
Mogge, line 28: Can this be done in some human readable way? Am I right, that we receive an ID, a slug, both or none from the params. So I suggest, if both (ID and slug) are undefined, return empty array or throw an error. We could say, that ID goes over slug if both are present.

Mogge (changes requested):
Looks like it works fine. Sorry for the drastic words, but the code to handle the params in the group resolver is a nightmare that nobody can read. It will be impossible to work on this code as it is in the future.

backend/src/schema/resolvers/groups.js
Tirokk: I couldn't find a good solution in Cypher.
What I have found was not working. So I spent only an hour. @Mogge 

I created two hopefully well named and therefore descriptional helper functions instead. Hope that satisfies you …

Mogge (approved):

PR #5489, opened 2022-10-15
refactor: 🍰 Disable Submit Button On group Update Changed Error

### LINKED ISSUE #5472: 🔧 [Refactor] Disable Submit Button On Update Formular If Nothing Has Changed
## 🔧 Refactor
when editing a group, the submit button will not work disable

ogerly:
## 🍰 Pull Request

when update edit a group, disable submit button if data are not according to schema

### Issues
- fixes #5472

### DIFF
--- a/webapp/components/Group/GroupForm.vue
+++ b/webapp/components/Group/GroupForm.vue
@@ -123,7 +123,7 @@
         <!-- location -->
         <ds-select
           id="city"
-          :label="$t('settings.data.labelCity')"
+          :label="$t('settings.data.labelCity') + locationNameLabelAddOnOldName"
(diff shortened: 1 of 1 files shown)


Tirokk (changes requested):
I have still some little problems:

- general:
  - the `clear location button` is not visible
- update:
  - if I change anything, but not the location, and save then the location vanishes on the group profile

I'm not sure this is all, but it seems the most imnportant to me.

ogerly (commented):
super, das problem scheint gelöst zu sein. ich habe keine fehler entdeckt

Tirokk (approved):
YES 😍🚀💫

PR #5544, opened 2022-10-22
change pixel mobile menu show by under 811px

ogerly:
## 🍰 Pullrequest

the header menu is empty when the width of the screen is exactly 810px. i have increased the maximum width to 811px, so that at 810px the mobile header is displayed normally.


ogerly:
is in https://github.com/Ocelot-Social-Community/Ocelot-Social/pull/5546 erledigt

Issue #5545, opened 2022-10-22
🚀 [Feature] Make Footer And Header Links Configurable If They Open A New Browser Window


Tirokk:
## 🚀 Feature

Make footer and header links configurable if they open a new browser window.

### Additional Informations

For rebranding purposes we like to have:

In files `webapp/layouts/default.vue` and `~/components/PageFooter/PageFooter.vue` …

- header menu links and the footer links …
  - … if they are configured for an external URL then …
    - … an additional configuration if the external URLs should be opened in a new browser window respectively tab

PR #5546, opened 2022-10-22
feat: 🍰 Header Menu To Component And Other Refinements

ogerly:
## 🍰 Pullrequest

I have packed the header menu from the _layout/default_ into a componente.
on the _layout/default_ ***maxMobileWidth: 811*** is determined in a central position and **showMobileMenu** is passed through to the menu.
the switching on and off of header mobil menu and footer is only controlled on the _layout/default_. i deleted all css that did that. so the default/layout is now very clear and we only have to change the value in one place if necessary.

### Issue

relates #5545


webapp/layouts/default.vue
```diff
@@ -217,65 +17,28 @@
...
-    },
     showMobileMenu() {
       if (!this.windowWidth) return false
-      return this.windowWidth < 810
-    },
-  },
-  methods: {
-    toggleMobileMenuView() {
-      this.toggleMobileMenu = !this.toggleMobileMenu
+      return this.windowWidth < this.maxMobileWidth
```
Tirokk, line 41:
```suggestion
      return this.windowWidth <= this.maxMobileWidth
```

That fits better to the name `maxMobileWidth` and the logic is more consitent if we use `810px` all over the place and not `811px`.

webapp/layouts/default.vue
```diff
@@ -217,65 +17,28 @@
...
       windowWidth: null,
-      links,
-      LOGOS,
-      isHeaderMenu: headerMenu.MENU.length > 0,
-      menu: headerMenu.MENU,
-      mobileSearchVisible: false,
-      toggleMobileMenu: false,
-      inviteRegistration: this.$env.INVITE_REGISTRATION === true, // for 'false' in .env INVITE_REGISTRATION is of type undefined and not(!) boolean false, because of internal handling,
-      categoriesActive: this.$env.CATEGORIES_ACTIVE,
+      maxMobileWidth: 811,
```
Tirokk, line 35:
```suggestion
      maxMobileWidth: 810,
```

See comment below.

Tirokk (changes requested):
An upfront review only because I've accidently have seen it.

Just let us change it next week with the other things which may appear …

Mogge (approved):

PR #5631, opened 2022-11-07
fix: My Groups Count Includes Pending Membership

### LINKED ISSUE #5628: 🐛 [Bug] Counter of my groups is incorrect due to pending memberships
## 🐛 Bugreport

Counter of my groups is incorrect due to pending memberships

Mogge:
### Issues
- fixes #5628 

### Todo

- [X] None

### DIFF
--- a/backend/src/schema/resolvers/groups.js
+++ b/backend/src/schema/resolvers/groups.js
@@ -97,7 +97,7 @@ export default {
         if (isMember) {
           cypher = `MATCH (user:User)-[membership:MEMBER_OF]->(group:Group)
                     WHERE user.id = $userId
-                    AND membership.role IN ['usual', 'admin', 'owner']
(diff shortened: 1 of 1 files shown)


backend/src/schema/resolvers/groups.js
```diff
@@ -97,7 +97,7 @@ export default {
         if (isMember) {
           cypher = `MATCH (user:User)-[membership:MEMBER_OF]->(group:Group)
                     WHERE user.id = $userId
-                    AND membership.role IN ['usual', 'admin', 'owner']
+                    AND membership.role IN ['usual', 'admin', 'owner', 'pending']
```
ulfgebhardt, line 100: is there no way to reduce the two queries to one, so this problem will never araise in the future anymore?

ulfgebhardt (approved):

PR #5683, opened 2022-11-16
fix(webapp): add text class settings for ellipsis text display

### LINKED ISSUE #5571: 🐛 [Bug] User Teaser Overflows The Post Teaser
## 🐛 Bugreport

User teaser overflows the post teaser.

Found by @hermann-san 🍀

photo_2022-10-24 11 56 51

SvFedoriv:
### Issues

- fixes #5571 

### Todo

- [X] None


SvFedoriv:
I've setted text elelement to `display:inline-block` and added width. That should solve the issue with ellipsis text display.
In case we want to display the text in the rows we could use `white-space: normal`.

ulfgebhardt:
@Tirokk can you lint this please?

PR #5707, opened 2022-11-24
fix(backend): sanitize group description

### LINKED ISSUE #5668: 🐛 [Bug][Backend, DB] Sanitize Group Properties `description`, `descriptionExcerpt` From Unwanted HTML Via `xssMiddleware`
## 🐛 Bugreport

In backend and db – sanitize group properties `description`, `descriptionExcerpt` from unwanted HTML via `xssMiddleware`.

See PR #5662 .
If you find a different solution then I would remove it from my PR.

### Additional Informations

We thought about a solution where we give @roschaefer `walkRecursive` an additional optional argument where we say on which GQL types we like to sanitize a property. @Tirokk @Mogge 

Old message from Discord:

```text
If we like to use `xssMiddleware` to sanitise HTML before it gets stored in the db – what we should –  we have to rename the `Group` node properties `description`.

This is because of the sanitising mechanism @roschaefer#5565 implemented which depends on the property names.
Because of the very general naming `description` this name is used in other places as well which should not be sanitised this way. That's the main reason. I haven't seen this from the beginning.

I made this a part of my issue:
https://github.com/Ocelot-Social-Community/Ocelot-Social/pull/5662

Therefore we need two db migrations, because it is not possible to do it in one:

– One for the full-text index.
– One for the properties.

I hope this all finds your agreement @Dornhoeschen#4105 @moriz#0232 ?

In the second migration I like to integrate a HTML sanitising like in `xssMiddleware`. That's because we have not sanitised HTML in the db in this properties.
Is this cool with you?
```

Mogge:
## 🍰 Pullrequest
Sanitizes the html of group description and description excerpt.
Additional argument to the walk recursive function to pass field names that should not be sanitized.

### Issues
- fixes #5668 

### Todo

- [X] None

### DIFF
--- a/backend/src/helpers/walkRecursive.js
+++ b/backend/src/helpers/walkRecursive.js
@@ -12,7 +12,7 @@ function walkRecursive(data, fields, fieldName, callback, _key) {
   if (data && typeof data === 'string' && fields.includes(_key)) {
     // well we found what we searched for, lets replace the value with our callback result
     const key = _key.split('!')
-    if (key.length === 1 || key[1] !== fieldName) data = callback(data, _key)
+    if (key.length === 1 || key[1] !== fieldName) data = callback(data, key[0])
   } else if (data && Array.isArray(data)) {
     // go into the rabbit hole and dig through that array
     data.forEach((res, index) => {
--- a/backend/src/helpers/walkRecursive.js
+++ b/backend/src/helpers/walkRecursive.js
@@ -2,24 +2,26 @@
  * iterate through all fields and replace it with the callback result
  * @property data Array
  * @property fields Array
+ * @property fieldName String
  * @property callback Function
  */
-function walkRecursive(data, fields, callback, _key) {
+function walkRecursive(data, fields, fieldName, callback, _key) {
   if (!Array.isArray(fields)) {
     throw new Error('please provide an fields array for the walkRecursive helper')
   }
   if (data && typeof data === 'string' && fields.includes(_key)) {
     // well we found what we searched for, lets replace the value with our callback result
-    data = callback(data, _key)
+    const key = _key.split('!')
+    if (key.length === 1 || key[1] !== fieldName) data = callback(data, _key)
   } else if (data && Array.isArray(data)) {
     // go into the rabbit hole and dig through that array
     data.forEach((res, index) => {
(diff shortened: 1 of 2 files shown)


backend/src/middleware/xssMiddleware.js
```diff
@@ -1,15 +1,22 @@
 import walkRecursive from '../helpers/walkRecursive'
 import { cleanHtml } from '../middleware/helpers/cleanHtml.js'
 
-const fields = ['content', 'contentExcerpt', 'reasonDescription']
+// exclamation mark separetes field names, that should not be sanitized
```
ulfgebhardt, line 4:
```suggestion
// exclamation mark marks field names, that should not be sanitized
```

backend/src/helpers/walkRecursive.js
```diff
@@ -2,24 +2,26 @@
...
-function walkRecursive(data, fields, callback, _key) {
+function walkRecursive(data, fields, fieldName, callback, _key) {
   if (!Array.isArray(fields)) {
     throw new Error('please provide an fields array for the walkRecursive helper')
   }
   if (data && typeof data === 'string' && fields.includes(_key)) {
     // well we found what we searched for, lets replace the value with our callback result
-    data = callback(data, _key)
+    const key = _key.split('!')
+    if (key.length === 1 || key[1] !== fieldName) data = callback(data, key[0])
```
ulfgebhardt, line 15: I do not understand `key[1] !== fieldName` shouldnt't it be `key[0] !== fieldName` or have `callback(data, key.length === 1 ? key[0] : key[1])

backend/src/middleware/xssMiddleware.js
Mogge: No. It does not mark, it separates.

backend/src/helpers/walkRecursive.js
Mogge: It is `description!embed`.

Mogge: And I want to pass `description` to the callback

Tirokk (approved):
Hey cool! @Mogge 

Thanks for caring! 🍀

I approve! 🚀🚀💫

PR #5973, opened 2023-02-15
fix(backend): flaky unit tests related to user avatar

Mogge:
## 🍰 Pullrequest
Hopefully eliminates the flakiness of the unit tests related to user avatar.

Number of successful consecutive runs: 1

### Issues

- None

### Todo

- [X] None


ulfgebhardt:
@Mogge  I give up on https://github.com/Ocelot-Social-Community/Ocelot-Social/pull/5973

We have two different Errors occuring: 
`Neode.get` expects an result, but the result is empty
- The Database reports it cannot find a certain Node
-> Both happens when trying to set a relation (since it requires two nodes to be present)

Possible Reasons:
- The Database is shit and causes this issue
- The Neo4J Driver causes this
- Neode causes the problem - a promise returned from neode for e.g. create does not ensure that the node actually is created in the database
- We have a problem with the jest tests - they allocate 4Gb memory and have ~200MB leak each suite -> This could be the reason things break down on the CI -> contraindicator: The error can happen very ealy in the testrun
- Rosie our factory library is the cause of the problem (unlikely - error occurs with an function and should not be linked to the library)
- The delete call `cleanDatabase` is not finished and it still deletes nodes while the next test does seed new data which is immediately - i checked for missing awaits and could not fix this problem

ulfgebhardt:
closing this for now

PR #5996, opened 2023-02-21
fix(backend): user or group has null as description

### LINKED ISSUE #5979: 🐛 [Bug] Map - users without descriptions have null on map shown
## 🐛 Bugreport

Map - users without descriptions have null on map shown

image

Elweyn:
## 🍰 Pullrequest

As user it is strange to see a description block with 'null' written in it.
Check that about !== 'null' cause the resolver has an undefinedToNull on about.

### Issues

- fixes #5979

### Todo

- [X] Fix the backend response


webapp/pages/map.vue
```diff
@@ -257,7 +257,9 @@ export default {
           </div>
            `
         description +=
-          e.features[0].properties.about && e.features[0].properties.about.length > 0
+          e.features[0].properties.about &&
+          e.features[0].properties.about.length > 0 &&
+          e.features[0].properties.about !== 'null'
```
Tirokk, line 262: How it commes that `e.features[0].properties.about` becomes a string with value 'null'.
That should not happen in my view.

So it should be this, but this is part of the first line and therefore superfluid:

```suggestion
          e.features[0].properties.about !== null
```

Tirokk (changes requested):
Very cool @Elweyn 😍

Would you like to change the seeding that we can accidently see that the null about works?
Instead of testing …

- to have two prominent user without an about:
  - about: null
  - about: ''
- to have at least one additional group without an about:
  - about: null
  - maybe additional – about: ''

Mogge (approved):

PR #6021, opened 2023-02-23
fix(webapp): remove additional markers from map

### LINKED ISSUE #5980: 🐛 [Bug] Map - search point has same color as groups
## 🐛 Bugreport

Map - search point has same color as groups

image

Elweyn:
## 🍰 Pullrequest

As user when I search an address in the Mapbox a blue marker appears but the blue marker is also used by groups.
After discussion with @Tirokk we agreed to remove additional markers.

### Issues

- fixes #5980 

### Todo

- [X] Remove additional marker posibiliter in MapboxGeocoder


Tirokk (approved):
Hey @Elweyn 

nice 👍🏼

I approve 🚀💫

PR #6088, opened 2023-03-06
fix(other): lint pr only on pullrequest

ulfgebhardt:
## 🍰 Pullrequest

lint pr only when there is a pullrequest not on every push

### Issues

- None

### Todo

- [X] None


Tirokk (approved):
Cool 😍

PR #6093, opened 2023-03-07
fix(webapp): error in user profil on the 'show x more' button 

### LINKED ISSUE #5967: 🐛 [Bug] Fix show more following users on user profile page
## 🐛 Bugreport

Clicking the `show x more...` button on the profile page leads to the following:
- more users are not displayed
- the spinning wheel keeps on spinning
- the profile picture disappears
- the display name shows Anonymous.
- the user slug shows @anonymous
- the member since value disappears
- the location value disappears
- the counter of followers and following shows 0
- the counter of posts, commented and shouted shows 0

Ocelot version 2.4.0
Firefox on Linux

Elweyn:
## 🍰 Pullrequest

When loading more user following or followed by we got an error.

### Issues

- fixes #5967 

### Todo

- [X] Change the value given to search


mahula (commented):
Well done.

Two issueare still to take care of:

- clicking the `show more` button: on both `show more` buttonsthe spinner spins
- the filter does not work with `@username`

Issue #6097, opened 2023-03-07
🚀 [Feature] Backend / Webapp – Group / Notification – Send User a Notification that their Rolle has been Changed


Tirokk:
## 🚀 Feature

Send User a Notification that their Rolle has been Changed.

```gherkin
Feature: Notification on member roll change
  As a user
  I want to be notified
  if I get a none pending membership roll in a group I entered
  or if I'm added to a group.
  Probably I like to be notified on all changes of my member roll
  as well if I enter a group or be added
  or if I get thrown out of a group or leave it.
```

PR #6144, opened 2023-03-15
feat(backend): notifications for join group

Mogge:
## 🍰 Pullrequest
Creates a notification for the owners of the group, when a user joins that group.

### Issues
- relates #6097 

### Todo

- [X] None

### DIFF
--- a/backend/src/schema/types/enum/ReasonNotification.gql
+++ b/backend/src/schema/types/enum/ReasonNotification.gql
@@ -2,4 +2,5 @@ enum ReasonNotification {
   mentioned_in_post
   mentioned_in_comment
   commented_on_post
+  user_joined_group
 }
--- a/backend/src/middleware/notifications/notificationsMiddleware.js
+++ b/backend/src/middleware/notifications/notificationsMiddleware.js
@@ -51,6 +51,19 @@ const publishNotifications = async (context, promises) => {
   })
 }
 
+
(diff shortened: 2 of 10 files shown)


backend/src/schema/types/enum/ReasonNotification.gql
Mogge, line 1: This file can be deleted?

Tirokk (changes requested):
In general this looks really good!

- in the screenshot we see that without a content the menu gets really small
- therefore and in general we may add the description of the group as the content?

Bildschirmfoto 2023-03-16 um 11 33 59


–––––––––––––––––––


PS: Sorry, I tested before I read the code …
So the following you hadn't already in your focus jet …

Scenario I tested:

### In „School For Citizens“:

- owners are Jenny and Louie
- Jenny thows Bob out

### Bobs Side

- notification for Bob is from Louie
  - should be from Jenny
- text is that Louie `joined`
  - should be that I'm (Bob) be `thrown out`

Bildschirmfoto 2023-03-16 um 10 02 31

### Jennies Side

- notification for Jenny is from Dewey
  - should be from Bob (or herself?)
- text is `joined`
  - should be `is removed`

Bildschirmfoto 2023-03-16 um 10 11 03

Tirokk:
close this in favor of https://github.com/Ocelot-Social-Community/Ocelot-Social/pull/6150

PR #6145, opened 2023-03-15
fix(other): use access token to trigger build on external repo

ulfgebhardt:
### Issues

- use access token to trigger build on external repo

### Todo

- [X] None


Tirokk (approved):
I approve 👍🏼

PR #6152, opened 2023-03-17
fix(webapp): join leave button for pending members

Mogge:
### Issues

- None

### Todo

- [X] None


Elweyn (approved):

Tirokk (changes requested):
Unfortunately there is no tooltip visible at the leave button named `Pending member` …

Not on FireFox nor on Safari.

Tirokk (approved):
Hey @mogge 

great! This works fine and as expected for me. 🍀👍🏼

I approve 🚀💫

PR #6167, opened 2023-03-20
fix(backend): new migration to create search indexes

ulfgebhardt:
## 🍰 Pullrequest

new migration to create search indexes. This reduces the role of the `db:migrate init` functionality and sets the fulltext search indexes reliable. Furthermore the index creating is now aligned with the group index.

This migration is special in the form that the indexes are dropped in case they already exist.

### Issues

- None

### Todo

- [X] None


Elweyn (approved):
For it seems good!
Just a question if on migration down we should also check if the index exists!?

Issue #6043, opened 2023-02-27
🚀 [Feature] Backend, DB – Add `Event` like `Post` for neo4j


Elweyn:
## 🚀 Feature
Create an Event as a special Post with additional informations, so we can use the already defined Post functions.
Extra informations:
- label "Event"
- location type
  - online event
  - geolocation
- start date and time
- end date and time

## Todo
- [x] Migration for all existing Posts to add additional label "Article"
- [x] Create Post Mutation needs a new Enum parameter to define if it creates an Article or an Event
  - [x] Add label "Post" and label ("Event" or "Article")
- [x] Update Post Mutation change second label "Event" to "Article" or "Article" to "Event"

PR #6197, opened 2023-03-29
feat(backend): create and update posts with labels

Mogge:
### Issues
- relates #6043 

### Todo

- [X] None

### DIFF
--- a/backend/src/schema/resolvers/posts.spec.js
+++ b/backend/src/schema/resolvers/posts.spec.js
@@ -4,6 +4,9 @@ import gql from 'graphql-tag'
 import { getNeode, getDriver } from '../../db/neo4j'
 import createServer from '../../server'
 import { createPostMutation } from '../../graphql/posts'
+import CONFIG from '../../config'
+
+CONFIG.CATEGORIES_ACTIVE = true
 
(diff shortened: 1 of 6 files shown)


backend/src/schema/resolvers/posts.spec.js
```diff
@@ -462,9 +496,25 @@ describe('UpdatePost', () => {
...
+              postType: ['Event'],
+            },
+          },
+          errors: undefined,
+        })
+      })
+    })
 
-    describe('params.image', () => {
+    describe.skip('params.image', () => {
```
Tirokk, line 517: It gives errors on my machine as well!

Why you skipped this test now?

It should have worked before?!

Tirokk (approved):
Hey @Mogge 

It looks really good in general! 😍

And I will approve. 🚀🚀💫💫

I have found:

- why it is neccessary to skip one test now?
- if I seed, not all posts `Post` (182) have label 'Article' (9), but there are no `Event` post still
  - so we have to fix the seeding I gues!?
- in GraphQL Playground I get sometimes the following error if I change the post type, see screenshot
  - has prpbobly nothing to do with you

PR #6242, opened 2023-04-20
fix(backend): group posts cannot be pinned

### LINKED ISSUE #6232: 🐛 [Bug] Backend - Pinned post of secret groups are also visible for users that are not member
## 🐛 Bugreport

As member or admin of a (secret, privat or public) group I want to be able to pin a post.
The actual function makes the post pinned on the overview and visible for every network user.
The optimal solution would be that as member of a secret or privat group I see the pinned posts in the overview and in the group as first posts!

## 🤖 Todo
- [ ] Remove in the getPosts query the pinned posts of privat and secret groups that I am not member

Mogge:
### Issues
- fixes #6232 

### Todo

- [X] None

### DIFF
--- a/backend/src/schema/resolvers/posts.js
+++ b/backend/src/schema/resolvers/posts.js
@@ -310,6 +310,7 @@ export default {
             `
             MATCH (user:User {id: $userId}) WHERE user.role = 'admin'
             MATCH (post:Post {id: $params.id})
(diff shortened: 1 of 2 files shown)


Elweyn (approved):
Test if admin can pin post in groups where he is member; doesn't work!
Good job!

Issue #6045, opened 2023-02-27
🚀 [Feature] Add `Event` Page that Shows Informations Specific for `Event`


Elweyn:
##  🚀 Feature

As a user I want to see specific Event Information on an event page.

View event page like post with additionals:

- meeting point:
  - details see #6044 
- start date-time
  - details see same
- optional: duration
  - details see same

Elweyn:
resolved through #6292

PR #6243, opened 2023-04-21
feat(webapp): [WIP] add event detail page

maeckes-infinitylabs:
## 🍰 Pullrequest

Adds an event route with event detail page as endpoint.

### Issues

- relates #6045 

### Todo

- [x] Add Event Details to PostDetail Component
- [ ] Add Map View for event location to PostDetail Component
- [ ] Translate new strings


Mogge:
@maeckes-infinitylabs I did rebase this PR on the `event-master` branch. I suggest, you work on this branch as base, to have the backend available for events.
I suggest a different approach: Do not create a event page. We use the same page for events and articles. Inside the post page, you check, if `postType === 'Event'` and render the `EventDetails` if true.

Mogge:
These are the new properties for posts you can expect from the backend:
```
  postType: [PostType]
    @cypher(statement: "RETURN filter(l IN labels(this) WHERE NOT l = 'Post')")

  eventLocationName: String
  eventLocation: Location @cypher(statement: "MATCH (this)-[:IS_IN]->(l:Location) RETURN l")
  eventVenue: String
  eventStart: String
  eventEnd: String
  eventIsOnline: Boolean
```

Mogge:
```
enum PostType {
  Article
  Event
}
```

Elweyn (changes requested):
See @Mogge comments

maeckes-infinitylabs:
> I did rebase this PR on the event-master branch. I suggest, you work on this branch as base, to have the backend available for events.

Thanks for the advice, will do.

> I suggest a different approach: Do not create a event page. We use the same page for events and articles. Inside the post page, you check, if postType === 'Event' and render the EventDetails if true.

I had the same idea but also added an event route:
- i created a postdetail vue component, that i added to the post route and also to the event route
- in the component itself i look which type it is and show details accordingly 

@Mogge 
i thought it would be good to have a specific event route. 
so should i revert this and just edit the post page and add not event route?

maeckes-infinitylabs:
> See @Mogge comments

thanks for the hint. just missed it :sweat:

PR #6256, opened 2023-04-25
fix(webapp): fix search for 3 chars

### LINKED ISSUE #6195: 🚀 [Feature] Backend, Webapp – Search Field – Start Search at `3` Characters
## 🚀 Feature

For a user it is strange to have a message 'Nothing found' after the third character and get results after the fourth character.

In search field start search at `3` characters.

Exclude functional characters like `!`, `@`, `&`, and `#`.

## 🤖 Todo
- [ ] Start search query after third character

ulfgebhardt:
## 🍰 Pullrequest

Remarks:
- Does not exclude functional characters as defined in the issue, but searches as soon as 3 characters are present (considering this negelectable)
- You can press enter even when there is one character which will result in you ending up on the search result page with the results for just one character search (considering this a feature)
- Also have the search refactored and simplified.

### Issues

- fixes #6195

### Todo

- [X] None

### DIFF
--- a/backend/src/schema/resolvers/searches.js
+++ b/backend/src/schema/resolvers/searches.js
@@ -223,7 +223,7 @@ export default {
     },
     searchResults: async (_parent, args, context, _resolveInfo) => {
       const { query, limit } = args
-      let userId = context.user?.id || null 
+      const userId = context.user?.id || null 
 
       const searchType = query.replace(/^([!@#&]?).*$/, '$1')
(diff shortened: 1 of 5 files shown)


Elweyn (dismissed):
The search starts at three character; special character @, ! and other should not be counted as search characters.
- If you just type 'hue' => Starts search ✅ 
- If you type '@hu' => Starts search ❌ Exepect to start at '@hue'

PR #6262, opened 2023-04-25
fix(backend): recover missing commit

Mogge:
### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/schema/types/type/Post.gql
+++ b/backend/src/schema/types/type/Post.gql
@@ -193,6 +193,7 @@ input _PostInput {
 
 input _EventInput {
   eventStart: String!
+  eventEnd: String
   eventLocation: String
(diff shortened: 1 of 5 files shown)


backend/src/schema/types/type/Post.gql
```diff
@@ -193,6 +195,7 @@ input _EventInput {
   eventStart: String!
   eventLocation: String
   eventVenue: String
+  eventIsOnline: Boolean
```
Elweyn, line 198: Don't we need eventEnd here?
```suggestion
  eventEnd: String
  eventIsOnline: Boolean
```

Elweyn (changes requested):

Elweyn (approved):

PR #6289, opened 2023-05-08
fix(webapp): properly render avatars in group settings

ulfgebhardt:
## 🍰 Pullrequest

This is the missing change for avatars - in the settings menu the avatars were missing.

### Issues

- None

### Todo

- [X] None


Elweyn (approved):
Works for me!

PR #6361, opened 2023-06-01
refactor(webapp): fix coverage

ulfgebhardt:
## 🍰 Pullrequest

Fixed coverage reporting. The coverage report is now properly generated using `@vue/vue2-jest` instead of `vue-jest`.

Additionally this defines `coverageProvider: "v8"` to get proper highlighting in the coverage html files.
See: https://jestjs.io/docs/configuration#coverageprovider-string

Apparently template and script elements are included in the coverage threshold and this leads to a high coverage value. Still this is an improvement since coverage now is correct(apparently was before as well) and is debugable via the html.

image

### Issues

- relates to https://github.com/vuejs/vue-jest/issues/522
- relates to https://github.com/vuejs/vue-jest/issues/480
- relates to https://github.com/vuejs/vue-test-utils/issues/1977

### Todo

- [x] Investigate if we can use babel as coverage provider to get more meaningful coverage values. -> **not working at the moment**


Mogge (changes requested):
`yarn test pages/search/search-results.spec.js `

`yarn test pages/post/edit/_id.spec.js`

Throw warnings. Please fix this

Elweyn (approved):

Issue #6336, opened 2023-05-25
🔧 [Refactor] Webapp: Create post page


Elweyn:
## 🔧 Refactor ticket

On the create "post" page we have some small design and text optimisation

## 🤖 Todo
- Two buttons "Create A New Post" & "Create A New Event"
  - DS-CARD-CONTENT: remove padding so we have no border around the button
  - Width takes more space than the form below
- Change text "Create a new Post" to "Create a new article"
  - Online Event
    - If check => hide "city / address" field

PR #6377, opened 2023-06-05
refactor(webapp): refactor create post page

maeckes-infinitylabs:
## 🍰 Pullrequest
On the Create Post Page:
- Changes text "Create a new Post" to "Create a new article"

- Removes padding so we have no border around the button when hovering. 
uses hover color of ds-button component:
image

- Fixes width issue of create form:
image

- Adds condition to hide event city if online-event is selected
image

### Issues

- relates #6336 

### Todo

- [x] Remove inline style and add css classes


Tirokk (changes requested):
Hey @maeckes-infinitylabs !

This works fine so far for me. 😍
Very cool !!!

I give one revert request for your locales.

PS: Before you merge just clearify: In case of hiding the location name be aware that @Elweyn experiments with wrapping the checkbox into a new component to validate it and maybe don't hide the location name, but make it not required.

Elweyn:
Has been taken over in new PR #6381

PR #6387, opened 2023-06-08
fix(webapp): remove filter by time from events

ulfgebhardt:
## 🍰 Pullrequest

remove filter by time from events.

This PR is a suggestion how to solve the Problem that events are missing when filtering for Events.

The solution of this PR is to remove additional time filter restrictions when filtering for Events. I believe that this is better understandable for the user.

Alternatively this can be fixed by a more complicated query, since how its currently implemented events that still happen are not displayed anymore when the start-date is in the past. The end-date is optional so that the filter must be set to the end-date if provided or start-date if not provided.

```js
TOGGLE_POST_TYPE(state, postType) {
    const filter = clone(state.filter)
    if (postType && !(filter.postType_in && filter.postType_in.includes(postType))) {
      filter.postType_in = [postType]
      /* if (postType === 'Event') {
        // check for `filter.eventEnd_gte = new Date()` if provided - else use the code below
        filter.eventStart_gte = new Date()
        state.order = 'eventStart_asc'
      } else {
        delete filter.eventStart_gte
        state.order = 'createdAt_desc'
      } */
    } else {
      delete filter.eventStart_gte
      delete filter.postType_in
      state.order = 'createdAt_desc'
    }
    state.filter = filter
  },
```

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/webapp/store/posts.js
+++ b/webapp/store/posts.js
@@ -86,13 +86,13 @@ export const mutations = {
     const filter = clone(state.filter)
     if (postType && !(filter.postType_in && filter.postType_in.includes(postType))) {
       filter.postType_in = [postType]
-      if (postType === 'Event') {
+      /* if (postType === 'Event') {
         filter.eventStart_gte = new Date()
         state.order = 'eventStart_asc'
       } else {
         delete filter.eventStart_gte
         state.order = 'createdAt_desc'
-      }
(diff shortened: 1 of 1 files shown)


Mogge:
Can we please close this. It is fixed in #6405

PR #6391, opened 2023-06-08
feat(backend): seed events

Mogge:
## 🍰 Pullrequest
 * sets the property `postType` to `Article` as default
 * seeds one active event
 * seeds one passed event
 * seeds one event started but not ended

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/db/seed.js
+++ b/backend/src/db/seed.js
@@ -596,7 +596,79 @@ const languages = ['de', 'en', 'es', 'fr', 'it', 'pt', 'pl']
       }),
     ])
 
-    // Create Posts
+    // Create Events (by peter lustig)
+    authenticatedUser = await peterLustig.toJson()
+    const now = new Date()
+
+    await Promise.all([
+      mutate({
+        mutation: createPostMutation(),
+        variables: {
+          id: 'e0',
+          title: 'Illegaler Kindergeburtstag',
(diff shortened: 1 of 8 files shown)


backend/src/db/seed.js
```diff
@@ -596,7 +596,79 @@ const languages = ['de', 'en', 'es', 'fr', 'it', 'pt', 'pl']
...
+        },
+      }),
+    ])
+
+    let passedEvent = await neode.find('Post', 'e1')
+    await passedEvent.update({ eventStart: new Date(2010, 8, 30, 10).toISOString() })
+    passedEvent = await neode.find('Post', 'e2')
+    await passedEvent.update({
+      eventStart: new Date(now.getFullYear(), now.getMonth(), now.getDate() - 3).toISOString(),
+    })
```
Tirokk, line 669: I do not understand, why you change the dates afterwards instead doing it in the first step?

PS: But you will have your thoughts and therefore it's no reason for me to not approve.

Tirokk (approved):
Year @Mogge ! 😍

Cool you write seedings.

I approve. 🚀💫💫

PS: I just have an understanding question.

backend/src/db/seed.js
Mogge: The API does not allow events in with a start date in the past. Therefore I have to pick a date in the future and then date it back

mahula (approved):
Well well

Issue #6439, opened 2023-06-13
🐛 [Bug] Yunite – Webapp – Registration – Terms and Condition + Data Privacy Links Going to Internal Pages


Tirokk:
## 🐛 Bugreport

Yunite – Webapp – Registration – Terms and Condition + Data Privacy Links Going to Internal Pages instead to the yunite.org link of the footer.

<img width="655" alt="Bildschirmfoto 2023-06-13 um 22 18 16" src="https://github.com/Ocelot-Social-Community/Ocelot-Social/assets/25344101/9930ca5d-f94e-4683-8544-b59013fb2e1b">

### Todo

- [x] prepare in main code: #6441
- [x] define in stage.yunite.me and yunite.me branding code:
  - [x] https://github.com/Yunite-Net/stage.yunite.me/commit/bb0574d733ab28c37ac65ab95190e6f3b709be97
  - [x] https://github.com/Yunite-Net/yunite.me/pull/3
  - [x] `https://yunite.me/terms-and-conditions` → `https://yunite.org/nutzungsbedingungen/`
  - [x] `https://yunite.me/data-privacy` → `https://yunite.org/datenschutz/`

Tirokk:
Solved by #6441 and in future by https://github.com/Yunite-Net/yunite.me/pull/3

PR #6441, opened 2023-06-14
fix(webapp): change registration `terms and conditions`, `data privacy` links to use `page-params-link`

Tirokk:
## 🍰 Pullrequest

change registration `terms and conditions`, `data privacy` links to use `page-params-link`

### Issues

- relates #6439

### Todo

- [x] force target `_blank`


webapp/components/_new/features/PageParamsLink/PageParamsLink.vue
```diff
@@ -1,15 +1,19 @@
...
   <a
     v-else
-    :href="pageParams.externalLink.url"
-    :target="pageParams.externalLink.target"
+    :href="
+      pageParams.isInternalPage ? pageParams.internalPage.pageRoute : pageParams.externalLink.url
+    "
+    :target="
+      forceTargetBlank ? '_blank' : pageParams.isInternalPage ? '' : pageParams.externalLink.target
+    "
```
Elweyn, line 16: I find it difficult to understand this and the whole component!

I would expect a component to be pretty blunt but now you set some logic in here, 
I think this complexity makes the reusability of the component pretty hard.

Elweyn: I would prefer to give the href & the target in the component as props and let the logic what is given to the props be done when calling it! (ex. webapp/components/Registration/RegistrationSlideCreate.vue)

Elweyn: pageParams would be a to wide definition.

Elweyn (dismissed):
Please look if we can't make the component PageParamsLink stupid.
If you need help you can ask me.

webapp/components/_new/features/PageParamsLink/PageParamsLink.vue
Tirokk: The component is used without logic all over the place, because it's always the same except in the new case.

With the name you are kind of right, but at that time I couldn't find an other short fitting name.

Please lets talk about it.

Elweyn (approved):

Issue #6392, opened 2023-06-08
🚀 [Feature] Backend: create a CRUD for rooms


Elweyn:
## 🚀 Feature

For the query of the list of rooms it would be great to have the room with the last message at first.

example code source: [https://github.com/antoine92190/vue-advanced-chat](https://github.com/antoine92190/vue-advanced-chat) 

```rooms="[
  {
    roomId: '1',
    roomName: 'Room 1',
    avatar: 'assets/imgs/people.png',
    unreadCount: 4,
    index: 3,
    lastMessage: {
      _id: 'xyz',
      content: 'Last message received',
      senderId: '1234',
      username: 'John Doe',
      timestamp: '10:20',
      saved: true,
      distributed: false,
      seen: false,
      new: true
    },
    users: [
      {
        _id: '1234',
        username: 'John Doe',
        avatar: 'assets/imgs/doe.png',
        status: {
          state: 'online',
          lastChanged: 'today, 14:30'
        }
      },
      {
        _id: '4321',
        username: 'John Snow',
        avatar: 'assets/imgs/snow.png',
        status: {
          state: 'offline',
          lastChanged: '14 July, 20:00'
        }
      }
    ],
    typingUsers: [ 4321 ]
  }
]"
````
## 🤖 TODOs
- [x] Mutation create a new room
- [x] Query list of rooms
  - the room with the last message at first
- [ ] Optional: Update room name
- [ ] Optional: Delete room

PR #6444, opened 2023-06-14
feat(backend): rooms

Mogge:
## 🍰 Pullrequest
creates and queries rooms

### Issues
- relates #6392 

### Todo

- [X] None


backend/src/schema/resolvers/rooms.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+})
+
+afterAll(async () => {
+  // await cleanDatabase()
+  driver.close()
+})
+
+describe('Room', () => {
+  beforeAll(async () => {
+    ;[chattingUser, otherChattingUser, notChattingUser] = await Promise.all([
```
ulfgebhardt, line 38: whats the semicolon for?

backend/src/schema/resolvers/rooms.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+      })
+    })
+
+    describe('authenticated', () => {
+      beforeAll(async () => {
+        authenticatedUser = await chattingUser.toJson()
+      })
+
+      describe('user id does not exist', () => {
+        it('returns null', async () => {
```
ulfgebhardt, line 79: i do not like that this returns `null` when its an error - but I assume thats because the query fails and returns null due to missing matching user.
```
MATCH (currentUser:User { id: $currentUserId })
MATCH (user:User { id: $userId })
```

One could consider to `if(!room) throw...`

backend/src/schema/resolvers/rooms.ts
```diff
@@ -0,0 +1,56 @@
...
+        id: context.user.id,
+      }
+      return neo4jgraphql(object, params, context, resolveInfo)
+    },
+  },
+  Mutation: {
+    CreateRoom: async (_parent, params, context, _resolveInfo) => {
+      const { userId } = params
+      const { user: { id: currentUserId } } = context
+      const roomId = uuid()
```
ulfgebhardt, line 19: This has collision problems.

Please use the database instead:
```
apoc.create.uuid()
```
Ref: https://neo4j.com/labs/apoc/4.1/graph-updates/uuid/

You would actually not require an extra id(the uuid), since every node has a `native ID` which has the disadvantage that it could be reused once the node was deleted. Therefore I suggest to use the uuid, but please with the database.

Further read:
https://stackoverflow.com/questions/60748609/neo4j-generate-unique-id-for-existing-node

backend/src/schema/resolvers/rooms.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+          await expect(mutate({
+            mutation: createRoomMutation(),
+            variables: {
+              userId: 'other-chatting-user',
+            },
+          })).resolves.toMatchObject({
+            errors: undefined,
+            data: {
+              CreateRoom: {
+                id: expect.any(String),
```
ulfgebhardt, line 105:
```suggestion
                id: expect.any(uuid),
```

backend/src/schema/resolvers/rooms.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+      
+      it('throws authorization error', async () => {
+        await expect(query({ query: roomQuery() })).resolves.toMatchObject({
+            errors: [{ message: 'Not Authorized!' }],
+          })
+      })
+    })
+
+    describe('authenticated', () => {
+      describe('as creater of room', () => {
```
ulfgebhardt, line 128:
```suggestion
      describe('as creator of the room', () => {
```

backend/src/schema/resolvers/rooms.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+          authenticatedUser = await chattingUser.toJson()
+        })
+
+        it('returns the room', async () => {
+          await expect(query({ query: roomQuery() })).resolves.toMatchObject({
+            errors: undefined,
+            data: {
+              Room: [
+                {
+                  id: expect.any(String),
```
ulfgebhardt, line 139:
```suggestion
                  id: expect.any(uuid),
```

backend/src/graphql/rooms.ts
```diff
@@ -0,0 +1,28 @@
...
+      $userId: ID!
+    ) {
+      CreateRoom(
+        userId: $userId
+      ) {
+        id
+      }
+    }
+  `
+}
```
ulfgebhardt, line 15: Why is this a function?

```suggestion
export const createRoomMutation = gql`
    mutation (
      $userId: ID!
    ) {
      CreateRoom(
        userId: $userId
      ) {
        id
      }
    }
  `
```

backend/src/graphql/rooms.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const roomQuery = () => {
```
ulfgebhardt, line 17: same

backend/src/schema/resolvers/rooms.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+          authenticatedUser = await otherChattingUser.toJson()
+        })
+
+        it('returns the room', async () => {
+          await expect(query({ query: roomQuery() })).resolves.toMatchObject({
+            errors: undefined,
+            data: {
+              Room: [
+                {
+                  id: expect.any(String),
```
ulfgebhardt, line 166:
```suggestion
                  id: expect.any(uuid),
```

backend/src/schema/resolvers/rooms.ts
```diff
@@ -0,0 +1,56 @@
...
+import { neo4jgraphql } from 'neo4j-graphql-js'
+import Resolver from './helpers/Resolver'
+
+export default {
+  Query: {
+    Room:  async (object, params, context, resolveInfo) => {
+      if (!params.filter) params.filter = {}
+      params.filter.users_some = {
+        id: context.user.id,
+      }
```
ulfgebhardt, line 11: This seems very hacky, but I assume thats how its suppose to be used?

You can hack this by providing the following filter:

```
_RoomFilter {
users_some: 'anything'
OR: _RoomFilter {
    users_some: 'my target user id'
  }
}
```
Remove the circular filters or the OR filter on `_RoomFilter`.

If you want to do it properly replace all `user_some` objects with the one defined in the code. Therfore you must walk the filter recursively. maybe loadash can provide an useful function

backend/src/schema/resolvers/rooms.ts
```diff
@@ -0,0 +1,56 @@
...
+          createRoomCypher,
+          { userId, currentUserId, roomId }
+        )
+        const [room] = await createRommTxResponse.records.map((record) =>
+                                                              record.get('room'),
+                                                             )
+        return room
+      })
+      try {
+        const room = await writeTxResultPromise
```
ulfgebhardt, line 40:
```suggestion
        const room = await writeTxResultPromise
        if(!room) throw Error
```

backend/src/schema/resolvers/rooms.ts
```diff
@@ -0,0 +1,56 @@
...
+      const writeTxResultPromise = session.writeTransaction(async (transaction) => {
+        const createRoomCypher = `
+          MATCH (currentUser:User { id: $currentUserId })
+          MATCH (user:User { id: $userId })
+          MERGE (currentUser)-[:CHATS_IN]->(room:Room)<-[:CHATS_IN]-(user)
+          SET room.createdAt = toString(datetime()),
+          room.id = $roomId
+          RETURN room { .* }
+        `
+        const createRommTxResponse = await await transaction.run(
```
ulfgebhardt, line 30: How does this compile?
```suggestion
        const createRommTxResponse = await transaction.run(
```

backend/src/schema/resolvers/rooms.ts
```diff
@@ -0,0 +1,56 @@
...
+      const { user: { id: currentUserId } } = context
+      const roomId = uuid()
+      const session = context.driver.session()
+      const writeTxResultPromise = session.writeTransaction(async (transaction) => {
+        const createRoomCypher = `
+          MATCH (currentUser:User { id: $currentUserId })
+          MATCH (user:User { id: $userId })
+          MERGE (currentUser)-[:CHATS_IN]->(room:Room)<-[:CHATS_IN]-(user)
+          SET room.createdAt = toString(datetime()),
+          room.id = $roomId
```
ulfgebhardt, line 27:
```suggestion
          ON CREATE SET
            room.createdAt = toString(datetime()),
            room.id = $roomId
```

backend/src/schema/resolvers/rooms.ts
```diff
@@ -0,0 +1,55 @@
...
+      const session = context.driver.session()
+      const writeTxResultPromise = session.writeTransaction(async (transaction) => {
+        const createRoomCypher = `
+          MATCH (currentUser:User { id: $currentUserId })
+          MATCH (user:User { id: $userId })
+          MERGE (currentUser)-[:CHATS_IN]->(room:Room)<-[:CHATS_IN]-(user)
+          ON CREATE SET
+          room.createdAt = toString(datetime()),
+          room.id = apoc.create.uuid()
+          RETURN room { .* }
```
ulfgebhardt, line 27:
```suggestion
          ON CREATE SET
            room.createdAt = toString(datetime()),
            room.id = apoc.create.uuid()
          RETURN room { .* }
```

ulfgebhardt (approved):

PR #6477, opened 2023-06-21
fix(backend): subscriptions

Mogge:
### Issues

- None

### Todo

- [X] None


backend/src/schema/resolvers/messages.spec.ts
```diff
@@ -234,32 +234,32 @@ describe('Message', () => {
             })).resolves.toMatchObject({
               errors: undefined,
               data: {
-                Message: [
-                  {
+                Message: expect.arrayContaining([
```
Mogge, line 237: This will be removed, when we implement the ordering of messages.

Mogge: The test as flaky

ulfgebhardt (approved):

PR #6530, opened 2023-07-11
fix(backend): chat - do not allow to create room with self

ulfgebhardt:
## 🍰 Pullrequest

chat - do not allow to create room with self

### Issues

- None

### Todo

- [X] None


ogerly (approved):

Elweyn (approved):

Elweyn (changes requested):
We need a Unit Test in Backend to not allow to create room with myself 😅

Elweyn (approved):

PR #6547, opened 2023-07-12
feat(backend): more message props

Mogge:
## 🍰 Pullrequest
add `distributed`, `saved` and `seen` props to messages.

### Issues

- None

### Todo

- [X] None


backend/src/schema/resolvers/messages.ts
```diff
@@ -58,6 +89,29 @@ export default {
         session.close()
       }
     },
+    MarkMessagesAsSeen: async (_parent, params, context, _resolveInfo) => {
+      const { messageIds } = params
+      const session = context.driver.session()
+      const writeTxResultPromise = session.writeTransaction(async (transaction) => {
+        const setSeenCypher = `
+          MATCH (m:Message) WHERE m.id IN $messageIds
```
ulfgebhardt, line 97: #must ensure only my messages are marked as seen, not those from others users
```
MATCH (user(myid)) - chats -> (room:Room) <- inside - (message(senderId != myid, id in messageIDs))
SET message.seen = true
```

ulfgebhardt: plus tests pls

ulfgebhardt (approved):

PR #6548, opened 2023-07-12
Feature(wepapp): changed color red to green.

### LINKED ISSUE #3086: 🚀 [Feature] Avoid the color red for non-essential things
> [alina-beck](https://github.com/alina-beck) **Authored by [alina-beck](https://github.com/alina-beck)**
_<time datetime="2020-02-17T16:15:05Z" title="Monday, February 17th 2020, 5:15:05 pm +01:00">Feb 17, 2020</time>_

---

## :rocket: Feature
requested by @DennisHack 

Red should be used very sparingly on the network – and only to focus our users' attention on the really important things, like charitable actions.

### User Problem

Users are distracted and their attention is guided towards the wrong details, due to our use of colors. Especially the notifications in the header should not be red as they are not as important as other features on the network.

#### Create Post

Bildschirmfoto 2023-03-27 um 15 29 03

### Implementation

1. go through the code/network and see where the color red is used
2. decide if it should be used in those locations (it is ok for destructive actions, like deleting a user account)
3. replace it with an alternative color – preferably green

### Design & Layout
See examples in our [clickdummy](https://preview.uxpin.com/24a2ab8adcd84f9a763d87ed27251351225e0ecd#/pages/99768901/simulate/sitemap?mode=i)

### Validation
Users will focus less on notifications.

entermex:
### Issues

- fixes #3086

### Additional Information

Looks now like:

Bildschirmfoto vom 2023-07-12 12-41-58
Bildschirmfoto vom 2023-07-12 12-42-03

### Todo

- [ ] still some

### DIFF
--- a/backend/src/graphql/messages.ts
+++ b/backend/src/graphql/messages.ts
@@ -16,10 +16,11 @@ export const createMessageMutation = () => {
 
 export const messageQuery = () => {
   return gql`
-    query ($roomId: ID!) {
-      Message(roomId: $roomId) {
+    query ($roomId: ID!, $first: Int, $offset: Int) {
+      Message(roomId: $roomId, first: $first, offset: $offset, orderBy: createdAt_desc) {
         _id
         id
+        indexId
         content
         senderId
         username
--- a/backend/src/schema/resolvers/messages.spec.ts
+++ b/backend/src/schema/resolvers/messages.spec.ts
@@ -215,6 +215,7 @@ describe('Message', () => {
                 {
                   id: expect.any(String),
                   _id: result.data.Message[0].id,
+                  indexId: 0,
                   content: 'Some nice message to other chatting user',
                   senderId: 'chatting-user',
                   username: 'Chatting User',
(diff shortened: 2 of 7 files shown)


webapp/assets/_new/styles/tokens.scss
```diff
@@ -15,7 +15,7 @@ $color-third: rgb(160,103,255);
 $color-success: rgb(23, 181, 63);
 $color-success-active: rgb(26, 203, 71);
 $color-success-inverse: rgb(241, 253, 244);
-$color-danger: rgb(219, 57, 36);
+$color-danger: rgb(25, 122, 49);
```
Tirokk, line 18: You can not just change the main color for dangerous things @entermex .

Because this danger color is used in the code somewhere else and should stay red.

webapp/locales/en.json
Tirokk, line 1: This is unrelated to your issue I guess.

Tirokk (changes requested):
Hey @entermex ,

the right solution form me is to remove the `danger` property from the related `base-button` tags in the file `webapp/components/ContributionForm/ContributionForm.vue`.

PR #6549, opened 2023-07-12
feat(webapp): chat message pagination

ulfgebhardt:
## 🍰 Pullrequest

This implements a working chat message pagination prototype.
Further more some stuff for the room query is done (unfinished).

### Issues

- None

### Todo

- [X] None


backend/src/schema/resolvers/messages.ts
```diff
@@ -13,11 +13,32 @@ export default {
...
+        const session = context.driver.session()
+        const countMessageTxPromise = session.readTransaction(async (transaction) => {
+          const countMessageCypher = `
+          MATCH((message:Message)-[:INSIDE]->(:Room { id: $roomId })) RETURN COUNT(message) AS count
+          `
+          const countMessageTxResponse = await transaction.run(countMessageCypher, {
+            roomId,
+          })
+          return await countMessageTxResponse.records[0].get('count')
         })
```
ulfgebhardt, line 30: All this crap could be removed if we could get an `indexId` with numeric unique incrementing value on the message type. It is unclear if there can be holes in the index id or not.

backend/src/schema/resolvers/messages.ts
```diff
@@ -13,11 +13,32 @@ export default {
...
-          message._id = message.id
+        const session = context.driver.session()
+        const countMessageTxPromise = session.readTransaction(async (transaction) => {
+          const countMessageCypher = `
+          MATCH((message:Message)-[:INSIDE]->(:Room { id: $roomId })) RETURN COUNT(message) AS count
+          `
+          const countMessageTxResponse = await transaction.run(countMessageCypher, {
+            roomId,
+          })
+          return await countMessageTxResponse.records[0].get('count')
```
ulfgebhardt, line 29: All this crap could be removed if we could get an `indexId` with numeric unique incrementing value on the message type. It is unclear if there can be holes in the index id or not.

ulfgebhardt: Fixed this

backend/src/schema/resolvers/messages.spec.ts
```diff
@@ -272,6 +275,72 @@ describe('Message', () => {
...
+                  roomId,
+                  first: 2,
+                  offset: 0,
+                  orderBy: 'created_desc',
+                },
+              }),
+            ).resolves.toMatchObject({
+              errors: undefined,
+              data: {
+                Message: expect.arrayContaining([
```
Mogge, line 304: Please remove array containing. The order must be correct as well

backend/src/schema/resolvers/messages.ts
```diff
@@ -13,7 +13,10 @@ export default {
           id: context.user.id,
         },
       }
+      // this does not work
+      // params.orderBy = ['createdAt_desc']
```
Mogge, line 17: remove

backend/src/schema/types/type/Message.gql
```diff
@@ -2,8 +2,14 @@
...
 # }
 
+enum _MessageOrdering {
+  createdAt_asc
+  createdAt_desc
+}
+
 type Message {
   id: ID!
+  indexId: String!
```
Mogge, line 12: Why not Int?

webapp/components/Chat/Chat.vue
```diff
@@ -147,20 +162,34 @@ export default {
...
             },
             fetchPolicy: 'no-cache',
           })
-          this.messages = Message
+
+          const msgs = []
+          ;[...this.messages, ...Message].forEach((m) => {
+            msgs[m.indexId] = m
+          })
+          this.messages = msgs.filter(Boolean)
```
Mogge, line 175: What is this?

Mogge (changes requested):

webapp/components/Chat/Chat.vue
ulfgebhardt: This filters out undefined indexes.
When you request message ids 35,36,37,38,39,40 and put them into an array the js engine seems to add the elements 0-34 as undefined. The library requires integer indexes, to properly work we need stable indexes, the order of the messages is defined by the index, newest messages have a higher ids.

The `map.filter(Boolean)` removes all undefined elements (all not truthy)

Mogge (approved):

PR #6554, opened 2023-07-13
refactor(other): cache docker images in github e2e flow

mahula:
## 🍰 Pullrequest
At the moment each e2e test job runs about 17 - 30 min.
Main reason is building the SUT docker images (about 15 min) in each e2e test job.

e2e_test_flow_before 

### Todo
- [x] extract docker image building into a separate preparation job to build only once
- [x] use [Github Cache Action](https://github.com/actions/cache) to store and acces the docker images
  - [x] delete cache after test runs

### Result
Now one preparation job builds images once. The e2e test jobs run 15 min less now.
This way the used runners will be freed much earlier for other workflow jobs.

   e2e_test_flow_after


.github/workflows/test-e2e.yml
```diff
@@ -1,9 +1,53 @@
...
+      - name: Get pr number
+        id: pr
+        uses: 8BitJonny/gh-get-current-pr@2.2.0
+      
+      - name: Cache docker images
+        id: cache
+        uses: actions/cache@v3.2.6
+        with:
+          path: /tmp/images/
+          key: e2e-preparation-cache-pr${{ steps.pr.outputs.number }}
```
ulfgebhardt, line 42: maybe you want to use the github action run instead to avoid cash misses? Not sure thats the reason for the full stack test currently failing, just a wild guess? 

image

`GITHUB_RUN_NUMBER` or `GITHUB_RUN_ID` https://docs.github.com/en/actions/learn-github-actions/variables

PR #6564, opened 2023-07-14
feat(backend): more room properties

Mogge:
## 🍰 Pullrequest
 * CreateMessage returns full message object
 * CreateRoom returns full room object
 * lastMessage property for room
 * unreadCount property for room

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/schema/resolvers/rooms.ts
+++ b/backend/src/schema/resolvers/rooms.ts
@@ -12,11 +12,6 @@ export default {
       if (resolved) {
         resolved.forEach((room) => {
           if (room.users) {
-            // buggy, you must query the username for this to function correctly
-            room.roomName = room.users.filter((user) => user.id !== context.user.id)[0].name
-            room.avatar =
-              room.users.filter((user) => user.id !== context.user.id)[0].avatar?.url ||
-              'default-avatar'
             room.users.forEach((user) => {
               user._id = user.id
             })
@@ -68,6 +63,7 @@ export default {
   },
   Room: {
     ...Resolver('Room', {
+      undefinedToNull: ['lastMessageAt'],
       hasMany: {
         users: '<-[:CHATS_IN]-(related:User)',
       },
--- a/backend/src/schema/resolvers/rooms.ts
+++ b/backend/src/schema/resolvers/rooms.ts
@@ -8,17 +8,7 @@ export default {
       params.filter.users_some = {
         id: context.user.id,
       }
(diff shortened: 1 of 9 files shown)


backend/src/schema/resolvers/rooms.ts
```diff
@@ -44,7 +29,14 @@ export default {
...
             room.createdAt = toString(datetime()),
             room.id = apoc.create.uuid()
-          RETURN room { .* }
+          WITH room, user, currentUser,
+          user.name AS roomName
+          RETURN room {
+            .*,
+            users: [properties(currentUser), properties(user)],
+            roomName: roomName,
+            unreadCount: toString(0)
```
ulfgebhardt, line 38: This `toString(0)` is not necessarily true, since you can (and will) call `createRoom` even for an existing room, which might contain unread messages for the user.

As I said before - the createRoom call has downsides (empty rooms), but for Implementation its easier. The rooms are queried after `createRoom` is called so we can certainly live with this little potential inaccuracy.

backend/src/schema/types/type/Room.gql
```diff
@@ -13,8 +13,23 @@ type Room {
...
-  roomName: String! ## @cypher(statement: "MATCH (this)<-[:CHATS_IN]-(user:User) WHERE NOT user.id = $cypherParams.currentUserId RETURN user[0].name")
-  avatar: String! ## @cypher match not own user in users array
+  roomName: String! @cypher(statement: "MATCH (this)<-[:CHATS_IN]-(user:User) WHERE NOT user.id = $cypherParams.currentUserId RETURN user.name")
+  avatar: String! @cypher(statement: "MATCH (this)<-[:CHATS_IN]-(user:User) WHERE NOT user.id = $cypherParams.currentUserId RETURN user.avatar.url")
+
+  lastMessageAt: String
+
+  lastMessage: Message @cypher(statement: """
+    MATCH (this)<-[:INSIDE]-(message:Message)
+    WITH message ORDER BY message.createdAt DESC LIMIT 1
```
ulfgebhardt, line 23:
```suggestion
    WITH message ORDER BY message.indexId DESC LIMIT 1
```

createdAt might be inaccurate even tho unlikely since it contains milliseconds.

backend/src/schema/resolvers/rooms.ts
Mogge: We need `toString` because as java number object is returned: `{ high: 0, low: 5 }. Javascript can not handle this object

ulfgebhardt (approved):

PR #6645, opened 2023-07-27
fix(webapp): add jsconfig.json to navigate in webapp.

Elweyn:
## 🍰 Pullrequest

In VsCode the Vue file could not be navigated through, 
with this PR it should be fixed.

### Issues

- None

### Todo

- [X] Development jump to imported vue files


webapp/jsconfig.json
```diff
@@ -0,0 +1,11 @@
+// tsconfig.json
```
mahula, line 1:
```suggestion
```

mahula (changes requested):

Elweyn:
@mahula OK let's check which one are working and which one need to be defined!

mahula (approved):

Issue #6634, opened 2023-07-24
🐛 [Bug][Webapp] Chat – `@` Mentions shall have correct link


Tirokk:

Tirokk:
Solved by #6646

PR #6646, opened 2023-07-28
fix(webapp): correct chat usertag profile link

maeckes-infinitylabs:
## 🍰 Pullrequest

When you now click on an user tag in a chat, it now redirects you correctly to the user profile page

### Issues

- #6634 

### Todo

- [X] None


Elweyn (approved):
Works as expected for me! 
We need to test the methods we have written! 
We should test everything in the Refactoring PR.

PR #6706, opened 2023-09-07
fix(other): add global package.json to workflow file filters

mahula:
## 🍰 Pullrequest
Changes in the global package.json might have effects on the whole system.
Thus all testing workflows should run completely to cover checking possible effects.


Elweyn (approved):
Thanks for the engagement!

PR #6720, opened 2023-09-11
fix(backend): fix post language detection

### LINKED ISSUE #6712: 🐛 [Bug] Backend - Setting post language fails, if no language was detected
## 🐛 Bugreport
Post creation/update includes setting the post's language.
Therefor a [language detection routine is called](https://github.com/Ocelot-Social-Community/Ocelot-Social/blob/92f179924f8d872e67c8b69b91b4b72917827e63/backend/src/middleware/languages/languages.ts#L7).

###  Actual behaviour
This call does not consider the case that the language detection returns an empty array - so it fails in this case.

### Expected behaviour
The [language setting routine](https://github.com/Ocelot-Social-Community/Ocelot-Social/blob/92f179924f8d872e67c8b69b91b4b72917827e63/backend/src/middleware/languages/languages.ts#L4) should provided a default language string (e.g. `en`).
This will be returned, if the language detection returns an empty array.

Tirokk:
- In case no language is detected set default.

## 🍰 Pullrequest

Fix post language detection.

### Issues

- fixes #6712

### Todo

- [X] None


backend/src/middleware/languages/languages.ts
```diff
@@ -1,20 +1,24 @@
...
 import { removeHtmlTags } from '../helpers/cleanHtml'
 
-const setPostLanguage = (text) => {
+const setPostLanguage = (text, defaultLanguage) => {
   const lngDetector = new LanguageDetect()
   lngDetector.setLanguageType('iso2')
-  return lngDetector.detect(removeHtmlTags(text), 1)[0][0]
+  let languages = lngDetector.detect(removeHtmlTags(text), 1)
+  if (!(Array.isArray(languages) && languages.length > 0)) {
+    languages = [[defaultLanguage, 1.0]]
```
Elweyn, line 9: I don't really understand the purpose of the ``1.0`` in this line but the rest seem good for me!

Elweyn (approved):

Mogge (approved):
Though I do not understand how language detection can fail. How can the text be empty?

PR #6704, opened 2023-09-05
fix(backend): metascraper crash

ulfgebhardt:
## 🍰 Pullrequest

Fixes a metascraper crash caused by `metascraper-audio`.

Further info see: https://github.com/microlinkhq/metascraper/issues/661

image

Explicit link: https://transinformation.net/offener-brief-an-den-widerstand-warum-lichtarbeiter-sich-gegenseitig-angreifen/

### Issues

- fixes #6703 

### Todo

- [X] None

PR #6724, opened 2023-09-12
refactor(backend): remove redundant package metascraper-audio from backend

mahula:
## 🍰 Pullrequest
In the fix #6704 usage of this package was removed from the backend code.
So it is redundant now.

### Issues
- relates #6704 #6632
- fixes #6632


Tirokk (approved):
I approve 👍🏼

PR #6752, opened 2023-09-25
fix(webapp): chat-only show counter if unread messages exist

### LINKED ISSUE #6750: 🚀 [Feature] Chat: Only show counter if unread messages exist
## 🚀 Feature

Currently, a counter 0 is shown if there are no unread chat messages. This is not needed and inconsistent with the logic of the notifications counter.

2023-09-23-unread-chat-counter-zero

**Solution**
Only show counter if there are any unread messages.

tarunsamanta2k20:
## 🍰 Pullrequest

Currently, a counter 0 is shown if there are no unread chat messages. This is not needed and inconsistent with the logic of the notifications counter.

2023-09-23-unread-chat-counter-zero
### Issues

- None
Closes #6750 

### Todo

- [X] None


webapp/components/_new/generic/CounterIcon/CounterIcon.vue
```diff
@@ -1,4 +1,4 @@
-<template>
+<template v-show="cappedCount > 0">
   <span class="counter-icon">
     <base-icon :name="icon" />
     <span :class="counterClass">{{ cappedCount }}</span>
```
Tirokk, line 4: Unfortunately this does not work, because the `v-show` is set at the `template`tag and because `cappedCount` is no number but a string.
And if the first would work then the whole icon would vanish and not only the number in the red circle.

Therefore I suggest:

```suggestion
<template>
  <span class="counter-icon">
    <base-icon :name="icon" />
    <span v-if="count > 0" :class="counterClass">{{ cappedCount }}</span>
```

Tirokk (changes requested):

Tirokk (approved):

Tirokk (approved):
Cool @tarunsamanta2k20 

Great that you have contribute to ocelot.social 😍
We really appriciate you piece of work!

I approve 🚀💫

PR #6784, opened 2023-10-09
fix(webapp): event date format

### LINKED ISSUE #6783: 🐛 [Bug] Events – On set Event Filter an Event of the same Day is not Visible
## 🐛 Bugreport

On set Event Filter an Event of the same Day is not Visible

### Additional Information

Created event:

Bildschirmfoto 2023-10-09 um 10 44 55

On filtering actual events the above event is missed:

Bildschirmfoto 2023-10-09 um 10 46 59

Mogge:
### Issues
- fixes #6783 

### Todo

- [x] migration to fix dates on server


Tirokk (commented):
Do we need a migration of the DB? @Mogge

Tirokk (commented):
That does work on Safari and Firefox. 👍🏼

Tirokk (changes requested):
- care for the summer and winter time @ulfgebhardt 
- give a backend error if the frontend sends not UTC datetime? @Mogge

backend/src/db/migrations/20231017141022-fix-event-dates.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export async function down(next) {
+  const driver = getDriver()
+  const session = driver.session()
+  const transaction = session.beginTransaction()
+
+  try {
+    // No sense in running this down
+    next()
```
ulfgebhardt, line 57: throw error?

webapp/components/ContributionForm/ContributionForm.vue
```diff
@@ -328,9 +328,9 @@ export default {
     eventInput() {
       if (this.createEvent) {
         return {
-          eventStart: this.formData.eventStart,
+          eventStart: new Date(this.formData.eventStart).toISOString(),
```
ulfgebhardt, line 331: How does this change the fact that we saved a timezoned Date?
Does this call ensure that the time ins transformed to UTC?

backend/src/db/migrations/20231017141022-fix-event-dates.ts
```diff
@@ -0,0 +1,68 @@
...
+to ISOString in UTC.
+`
+
+export async function up(next) {
+  const driver = getDriver()
+  const session = driver.session()
+  const transaction = session.beginTransaction()
+
+  try {
+    const events = await transaction.run(`
```
ulfgebhardt, line 14: To make this universal we could to the following:

```
if (post.createdAt is before this migration date) {// we specify a date here
   // We assume German time as timezone
   if(post.eventStart is German Summer Time){
      post.eventStart -=2h
      post.eventEnd -= 2h
   } else {
      post.eventStart -=1h
      ...
   }
}
```

You can use https://stackoverflow.com/questions/11887934/how-to-check-if-dst-daylight-saving-time-is-in-effect-and-if-so-the-offset to determine if DST is in effect.

ulfgebhardt (commented):
I think we should define a time before this transformation happens and one after this will not happen.
`if(post.createdAt is before XXX)`

If we should do the extra effort to calculate the summer and winter offset is not important to me.

I do not fully understand if the problem is solved with the Conversion to Date - i assume it takes the given timezone and converts it to UTC, but thats an assumption.

backend/src/db/migrations/20231017141022-fix-event-dates.ts
Tirokk: I ment throwing an error on creation and update the event so that we get aware and implement a practice to not allow wrong date formats send to the backend and to went into the database. @ulfgebhardt

webapp/components/ContributionForm/ContributionForm.vue
Mogge: Yes.

Mogge: The browser takes care of it.

backend/src/db/migrations/20231017141022-fix-event-dates.ts
Mogge: Really needed? I think not

Tirokk (approved):
Works excelent. @Mogge 

Therefore I approve 🚀🚀💫💫

PS: I tested the error message in the browser as well. Cool! 😍
PPS: And you already corrected the hour offset. 👍🏼

Tirokk:
**_!!! Before we merge we have to prepare the stage.yunite.me to test the migration on the server !!!_**

Tirokk (approved):
A very good idea to take the migration out of the coverage. 👍🏼

Tirokk:
**_!!! Before we merge we have to prepare the stage.yunite.me to test the migration on the server !!!_**

PR #6792, opened 2023-10-14
fix(backend): 6791 fix scripts to typescript

Tirokk:
### Issues

- fixes #6792

### Todo

- [x] in `backend/package.json`:
  - [x] `prod:migrate`
  - [x] `start`
- [x] in `deployment/src/kubernetes/README.md`
  - [x] see: `reset and seed Neo4j database via backend for staging`
- [x] add the following description of …

… clear the database with migrations, indexes, and contraints:

```text
Sonst alle noch einmal löschen:
https://github.com/Ocelot-Social-Community/Ocelot-Social/blob/master/neo4j/README.md

Datenbank leeren, bis auf Migrations:
Backend: `yarn run db:reset`
see: drop all indexes and constraints:
Neo4j Cypher: `CALL apoc.schema.assert({},{},true) YIELD label, key RETURN * ;`

Migrations löschen:
Neo4j Cypher: `MATCH (n:Migration) DETACH DELETE n RETURN n;`

Dann noch mal die Indexes neu setzen:
Backend: `yarn prod:migrate init`
Backend: `yarn run db:migrate up`
Und noch mal mit Testdaten bestücken.
Backend: `yarn run db:reset`
Backend: `yarn run db:seed`
```


Tirokk (commented):
How we can get this line running in TypeScript again?
@ulfgebhardt @Mogge 

https://github.com/Ocelot-Social-Community/Ocelot-Social/blob/ac8b3e3c2b281b12eee447ec4df9e63e89024d5c/deployment/src/kubernetes/README.md?plain=1#L296

This is only for production purposes.

I've seen that @ulfgebhardt  hasen't changed it in the deployment script `deployment/scripts/cluster.reseed.sh` and this isused in our deployment repo here:

https://github.com/Ocelot-Social-Community/stage.ocelot.social/blob/fdc2e52fa444b300e1c4736600bc0e9ae3314222/.github/workflows/deploy.yml#L57

Therefore it seems to work. But why it doesn't work locally?
As:

```bash
$ /bin/sh -c "node --experimental-repl-await build/src/db/clean.js && node --experimental-repl-await build/src/db/seed.js"
```

We would need it if we try production locally.

ulfgebhardt (changes requested):
I don't think this is the right solution.
The command `yarn prod:migrate` is to migrate in a production environment where a typescript compiler is unavailable. Therefore we cannot change this the way its suggested here.

To test: log into a running compiled production docker with exec and run the command `yarn prod:migrate init/up`. This must work!

Tirokk:
Is superfluid now

PR #6813, opened 2023-11-02
chore(other): add deployment script 'clusters.cron-backups.sh'

### LINKED ISSUE #6812: 💥 [DevOps] Script for Automated Backups for Kubernetes Networks
## 💥 DevOps ticket

Script for Automated Backups for Kubernetes Networks.

Tirokk:
## 🍰 Pullrequest

Add deployment script 'clusters.cron-backups.sh'.

### Issues

- fixes #6812

### Todo

- [x] delete oldest backups
- [x] consider to use an .env
- [x] cleanup the scripts from debugging code
- [x] merge #6893 
- [x] readme
  - [x] implement crontab
- [x] remove database name from scripts by set default database name in Docker files and Helm scripts, see #6894 
  - [ ] there is a problem with the database name on different clusters, not knowing why are differences
    - on `freilernen.social` the database name is:
      - `graph.db`
    - on `sender.fm`, `yunite.me` the database name is:
      - `neo4j`


deployment/scripts/clusters.cron-backups.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+# base setup
+SCRIPT_PATH=$(realpath $0)
+SCRIPT_DIR=$(dirname $SCRIPT_PATH)
+
+# check BACKUP_CONFIGURATIONS
+if [ -z ${BACKUP_CONFIGURATIONS} ]; then
```
ulfgebhardt, line 8: consider to use an .env

deployment/scripts/clusters.backup-multiple-servers.sh
```diff
@@ -0,0 +1,93 @@
...
+  # convert configurations to array
+  IFS=' ' read -a CONFIGURATIONS_ARRAY <<< "$BACKUP_CONFIGURATIONS"
+
+  # display the clusters
+  printf "Backup the clusters:\n"
+  for i in "${CONFIGURATIONS_ARRAY[@]}"
+  do
+    echo "  $i"
+  done
+fi
```
ulfgebhardt, line 34: You can remove the else block in most cases since you exit in the if block. The mechnic should be:

```
if(error_condition1){
  exit
}
continue execution command1
continue execution command2
continue execution command3
...
if(error_condition2){
  exit
}
...
```

deployment/scripts/clusters.backup-multiple-servers.sh
```diff
@@ -0,0 +1,93 @@
...
+
+    keep=$BACKUP_SAVED_BACKUPS_NUMBER
+    path="$SCRIPT_DIR/../configurations/$CONFIGURATION/backup/"
+
+    cd $path
+
+    printf "In\n  '$path'\n  remove:\n"
+    while [ `ls -1  |  wc -l` -gt $keep ]; do
+      # TODO: because 'ls' is not always relyable the same on different shells maybe replace 'ls' by 'find . -type d -maxdepth 1'? description: https://unix.stackexchange.com/questions/28939/how-to-delete-the-oldest-directory-in-a-given-directory
+      #         I tested this, but 'find' is crutial to use, because of shell compatibilities
```
ulfgebhardt, line 74: fix it, remove comment

deployment/scripts/clusters.backup-multiple-servers.sh
```diff
@@ -0,0 +1,93 @@
...
+  #%! echo "You must provide a BACKUP_SAVED_BACKUPS_NUMBER via environment variable"
+  printf "!!! You must provide a BACKUP_SAVED_BACKUPS_NUMBER via environment variable !!!\n"
+  exit 1
+else
+  # deleting backups?
+  if (( BACKUP_SAVED_BACKUPS_NUMBER >= 1 )); then
+    printf "Keep the last %d backups for all networks.\n" $BACKUP_SAVED_BACKUPS_NUMBER
+  else
+    echo "!!! ATTENTION: No backups are deleted !!!"
+  fi
```
ulfgebhardt, line 47: You seem to do this twice - its below in the loop again

deployment/scripts/clusters.backup-multiple-servers.sh
```diff
@@ -0,0 +1,93 @@
...
+      rm -rf $oldest
+    done
+
+    printf "Keep the last %d backups:\n" $BACKUP_SAVED_BACKUPS_NUMBER
+    ls -c1  |  sort -n  |  awk '{print "  " $0}'
+
+    cd $SCRIPT_DIR
+  else
+    echo "!!! ATTENTION: No backups are deleted !!!"
+  fi
```
ulfgebhardt, line 86: remove else - you have that above

Tirokk: Good catch.

But it was intentional and I like to keep this.

deployment/scripts/clusters.backup-multiple-servers.sh
Tirokk: Done

deployment/scripts/clusters.backup-multiple-servers.sh
Tirokk: Done for both cases

deployment/scripts/clusters.cron-backups.sh
Tirokk: Done

deployment/scripts/cluster.backup.sh
```diff
@@ -24,7 +29,7 @@ ${SCRIPT_DIR}/cluster.neo4j.sh maintenance on
 # database backup
 kubectl --kubeconfig=${KUBECONFIG} -n default exec -it \
     $(kubectl --kubeconfig=${KUBECONFIG} -n default get pods | grep ocelot-neo4j | awk '{ print $1 }') \
-    -- neo4j-admin dump --database=graph.db --to=/var/lib/neo4j/$BACKUP_DATE-neo4j-dump
+    -- neo4j-admin dump --database=neo4j --to=/var/lib/neo4j/$BACKUP_DATE-neo4j-dump
```
Tirokk, line 32: To be discussed, because @ulfgebhardt gave the hint that the changed db name `neo4j` come by loading dumps to `sender.fm` and `yunite.me`.

That seems plausable to me!

Therefore he suggests to set the default db name in the deployment of the db and then we can remove the `--database=graph.db` option.

So inverstigate this possibility, because it may be that this is only possible since version Neo4j v5 and not on our v4.4.

Tirokk: I find that you cannot change the database default name in Neo4j community edition, but only in enterprise edition:

- see https://neo4j.com/docs/operations-manual/4.4/reference/procedures/#_list_of_procedures: `dbms.cluster.setDefaultDatabase()`

ulfgebhardt: Ok this database is a very sad story ;)

If you want to solve the problem on database level(what I still recommend since it reduces code-impact) you can rename the databases on all those clusters still using `graph.db` as database name.
(`graph.db` is the neo4j 3.x database name and should not occur on any future installations)

Sadly the database does not provide a rename database command (as far as my research goes), but using dump & restore can solve this issue in no-time.

```
Backup the existing database into data/dumps folder by running this command in the terminal window:

bin/neo4j-admin backup --backup-dir=data/dumps --database=xyz

Restore this backup as a different database:

bin/neo4j-admin restore --from=data/dumps/xyz --database=xyzz --force
```
[Source](https://community.neo4j.com/t/creating-a-database-and-then-renaming-it-or-copying-it-to-another-within-the-same-instance/36374)

If you want to use the default database name you have to do the following:
1. remove all `--database=graph.db` from all commands, do not replace it with `--database=neo4j `, just remove it
2. All databases which are still imported and have the name `graph.db` -> Log into cluster and dump+restore database as described above with default name (`neo4j`).
3. ...
4. Profit

The option you suggested is still valid, just has way more implications
- make database name configurable
- configure it for every cluster accordingly
- adjust all scripts
- ensure proper env variable loading and passing through scripts
- ...

ulfgebhardt: Oh and here is also a copy command: https://neo4j.com/docs/operations-manual/4.4/backup-restore/copy-database/ lots of red text tho.

Tirokk: I find the solution to set the default database by configuration by use `NEO4J_dbms_default__database` as an environment variable when starting.

Tried that out and works.

Thanks for your help @ulfgebhardt !

Tirokk: I have removed `--database=`

deployment/scripts/cluster.maintenance.sh
```diff
@@ -5,7 +5,7 @@ SCRIPT_PATH=$(realpath $0)
 SCRIPT_DIR=$(dirname $SCRIPT_PATH)
 
 # check CONFIGURATION
-if [ -z ${CONFIGURATION} ]; then
+if [[ -z "$CONFIGURATION" ]] || [[ $CONFIGURATION == "" ]]; then
```
ulfgebhardt, line 8:
```suggestion
if [[ -z "$CONFIGURATION" ]]; then
```

-z "$CONFIGURATION" should check for empty string, the second condition is implicitly checked by the first.
Tho the change from `-z ${CONFIGURATION}` to `-z "$CONFIGURATION"` introduces this check, I think a recheck if the second condition is needed should be done: https://linuxsimply.com/bash-scripting-tutorial/conditional-statements/if-else/if-string-not-empty/

deployment/scripts/cluster.neo4j.sh
```diff
@@ -5,7 +5,7 @@ SCRIPT_PATH=$(realpath $0)
 SCRIPT_DIR=$(dirname $SCRIPT_PATH)
 
 # check CONFIGURATION
-if [ -z ${CONFIGURATION} ]; then
+if [[ -z "$CONFIGURATION" ]] || [[ $CONFIGURATION == "" ]]; then
```
ulfgebhardt, line 8: same

deployment/scripts/clusters.backup-multiple-servers.sh
```diff
@@ -0,0 +1,91 @@
...
+# save old CONFIGURATION for later reset
+export SAVE_CONFIGURATION=$CONFIGURATION
+
+# export all variables in "../.env"
+set -a            
+source ${SCRIPT_DIR}/../.env
+set +a
+
+# check BACKUP_CONFIGURATIONS
+if [[ -z "$BACKUP_CONFIGURATIONS" ]] || [[ $BACKUP_CONFIGURATIONS == "" ]]; then
```
ulfgebhardt, line 20: same

docker-compose.yml
```diff
@@ -111,10 +111,12 @@ services:
       # settings reference: https://neo4j.com/docs/operations-manual/4.4/docker/ref-settings/
       # TODO: This sounds scary for a production environment
       - NEO4J_AUTH=none
+      # - NEO4J_dbms_default__database=graph.db
```
ulfgebhardt, line 114: this works?

Tirokk: This works like we need it.
I went with greate energy to find out how we can achive the solution you sugested in an easy way.

If you set this name the dump command uses this database as default if you have no name given in the command.

If this database doesn't exist on starting Neo4j a database with this name gets created and is empty.
The other databases are not effected and you can use them again by stopping Neo4j and restart it with the database name you like to use.

I tested it locally very careful, because we don't want to make any harm.

If you like @ulfgebhardt lets merge this?

deployment/scripts/cluster.maintenance.sh
Tirokk: Your are right @ulfgebhardt I checked this and it works in all cirumstances: 'undefined' and ""

Therefore I've overtaken your suggestion in the code at all relevant places with co-authored.

deployment/scripts/cluster.neo4j.sh
Tirokk: done

deployment/scripts/clusters.backup-multiple-servers.sh
Tirokk: done

deployment/src/kubernetes/templates/neo4j/ConfigMap.yml
```diff
@@ -20,4 +20,5 @@ data:
   NEO4J_dbms_security_procedures_unrestricted: "{{ .Values.NEO4J.DBMS_SECURITY_PROCEDURES_UNRESTRICTED }}"
   NEO4J_dbms_allow__format__migration: "true"
   NEO4J_dbms_allow__upgrade: "true"
-  NEO4J_apoc_import_file_enabled: "{{ .Values.NEO4J.APOC_IMPORT_FILE_ENABLED }}" 
\ No newline at end of file
+  NEO4J_dbms_default__database: "{{ .Values.NEO4J.DBMS_DEFAULT_DATABASE }}"
```
Tirokk, line 23: I've added the database name setting to the Helm script

deployment/scripts/clusters.backup-multiple-servers.sh
```diff
@@ -0,0 +1,91 @@
...
+
+# check BACKUP_CONFIGURATIONS
+if [[ -z "$BACKUP_CONFIGURATIONS" ]]; then
+  #%! echo "You must provide a BACKUP_CONFIGURATIONS via environment variable"
+  printf "!!! You must provide a BACKUP_CONFIGURATIONS via environment variable !!!\n"
+  exit 1
+fi
+
+# check BACKUP_SAVED_BACKUPS_NUMBER
+if [[ -z ${BACKUP_SAVED_BACKUPS_NUMBER} ]]; then
```
ulfgebhardt, line 27: +

ulfgebhardt (approved):

PR #6822, opened 2023-11-04
refactor(other): set dependabot to assign version bump pr to devops user

mahula:
## 🍰 Pullrequest
Configure dependabot to assign its version bump PRs to @mahula

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,8 @@ updates:
     day: "saturday"
     timezone: "Europe/Berlin"
     time: "03:00"
+  assignees:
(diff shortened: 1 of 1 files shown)


ulfgebhardt (commented):
Why is this needed?
I believe this is a bad idea, since it assumes that you will take care of it forever.
Do you gain advantage of the assignment e.g. an notification you use?

mahula:
> Why is this needed? I believe this is a bad idea, since it assumes that you will take care of it forever. Do you gain advantage of the assignment e.g. an notification you use?

Good point.
No, this is not to important.

Issue #6952, opened 2024-01-24
💥 [DevOps] Refine script `yarn locales --fix`


Tirokk:
## 💥 DevOps ticket

Refine script `yarn locales --fix`.

### Todo

In secondary languages

- add `null` entry if entry is missed
- remove entry if entry is removed in primary language

Mogge:
The good first issue tag is a joke, isn't it?

PR #7038, opened 2024-03-05
feat(webapp): normalize locale files

Mogge:
## 🍰 Pullrequest

`yarn locales:normalize` added to scripts.
Copies the keys of the English locale file and to all other locale files (except German). If a value is present, it is set, else null.

We should discuss how this could be integrated in the CI. 

### Issues
- relates #6952 

### Todo

- [X] None

### DIFF
--- a/scripts/translations/normalize-locales.jq
+++ b/scripts/translations/normalize-locales.jq
@@ -1,7 +1,7 @@
 # normalize-locales.jq
 #
 # creates a copy of the structure of source and replaces the values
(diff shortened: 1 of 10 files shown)


Elweyn (approved):
Thanks for your work!
Just one question came up my mind, don't we want to have the normalize script in ``yarn locales``

Tirokk:
Lets do it in `yarn locales --fix` only? @Mogge @Mogge

PR #7144, opened 2024-03-28
feat(frontend): implement ocelot color tokens in frontend with themes

### LINKED ISSUE #7114: 🚀 [Feature][Frontend] Set Ocelot Color Tokens in Frontend
## 🚀 Feature

Set Ocelot Color Tokens in Frontend.

### Additional Information

- https://stackoverflow.com/questions/74374571/can-i-pass-css-variables-to-vuetify-theme-colors

Tirokk:
### Issues

- fixes #7114

### Todo

- [x] frontend runs, but in the unit test we get the warning `[Vue warn]: Vuetify: '_colorPrimary_2c143d' is not a valid hex(a) color`
  - [x] console.log gives `tokens.colorPrimary:  _colorPrimary_2c143d`
  - [x] can only be solved if in the theme the color is set directly to `primary: 'rgb(23, 181, 63)',` 
  - [x] recherche by searching for `vitest scss color variables warning vuetify 3` give no helpful results to me
  - [x] ~~lets make a separate issue~~ its fixed
- [ ] background color is not applied correct in theme `dark`. it's still kind of white
  - [x] lets make a separate issue and we will find out why if we go on: #7371


frontend/vitest.config.ts
```diff
@@ -28,7 +28,7 @@ export default mergeConfig(
         thresholds: {
           lines: 84,
           // functions: 20, // has problems see https://github.com/vitest-dev/vitest/issues/3607
-          branches: 92,
+          branches: 89,
```
Bettelstab, line 31: Why do we need to change this here?

Tirokk: The lines of code have increased mainly by adding the file `tokens.module.scss` but as far as I see there is no sense to test the lines. Therefore the relation of tested lines to total count of lines decreases.

Bettelstab (approved):

frontend/vitest.config.ts
ulfgebhardt: This is not true

image

This part is uncovered and represents new functionality, namely a button, which is untested. This is correctly detected.

I think its valid to reduce coverage tho, just correcting the statement.

image

Tirokk: Sorry for my wrong statement and thanks for correcting me @ulfgebhardt

Tirokk:
Thanks a lot to fix the CSS warning in the tests! 🥰 @ulfgebhardt 

I created an issue for the problem to assign a black app background in the `dark` theme: #7371 
Can we merge this PR and go on? @Bettelstab @ulfgebhardt @Elweyn @mahula @maeckes-infinitylabs

Elweyn (approved):

PR #7848, opened 2024-09-17
chore(other): adjust docs for deployment, neo4j cypher commands, and have new scripts

ulfgebhardt:
## 🍰 Pullrequest

Adjust docs for deployment, neo4j cypher commands, and have new scripts

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/deployment/scripts/cluster.backend-bash.sh
+++ b/deployment/scripts/cluster.backend-bash.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# time stamp
+printf "Neo4J bash :\n  "
+date
+
+# base setup
+SCRIPT_PATH=$(realpath $0)
+SCRIPT_DIR=$(dirname $SCRIPT_PATH)
+
+# check CONFIGURATION
+if [[ -z "$CONFIGURATION" ]]; then
+  echo "!!! You must provide a CONFIGURATION via environment variable !!!"
+  exit 1
+fi
+
+printf "  Cluster:  %s\n" $CONFIGURATION
+
+# configuration
+KUBECONFIG=${KUBECONFIG:-${SCRIPT_DIR}/../configurations/${CONFIGURATION}/kubeconfig.yaml}
+
+kubectl --kubeconfig=${KUBECONFIG} -n default exec -it $(kubectl --kubeconfig=${KUBECONFIG} -n default get pods | grep ocelot-backend | awk '{ print $1 }') -- /bin/sh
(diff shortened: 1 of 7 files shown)


Tirokk (dismissed):

deployment/scripts/cluster.dashboard-tunnel.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
roschaefer, line 1: I would say just use `k9s` and drop the kubernetes dashboard. Or use both and use `k9s` to set up the tunnel for you.

deployment/scripts/cluster.backend-bash.sh
```diff
@@ -0,0 +1,22 @@
...
+  echo "!!! You must provide a CONFIGURATION via environment variable !!!"
+  exit 1
+fi
+
+printf "  Cluster:  %s\n" $CONFIGURATION
+
+# configuration
+KUBECONFIG=${KUBECONFIG:-${SCRIPT_DIR}/../configurations/${CONFIGURATION}/kubeconfig.yaml}
+
+kubectl --kubeconfig=${KUBECONFIG} -n default exec -it $(kubectl --kubeconfig=${KUBECONFIG} -n default get pods | grep ocelot-backend | awk '{ print $1 }') -- /bin/sh
```
roschaefer, line 22: This is to SSH into the backend? It's easier accomplished with `k9s`.

roschaefer (changes requested):
I would say let's remove the bash scripts, no?

deployment/scripts/cluster.dashboard-tunnel.sh
Tirokk: Thanks for the hint @roschaefer !
I get your point and we will change this in the future. 💫

Please just leave it that way for now and we will modernize all these in the next step.
Cool?

deployment/scripts/cluster.backend-bash.sh
Tirokk: Yes, this is to SSH into the backend.

same same as obove?

deployment/scripts/cluster.dashboard-tunnel.sh
roschaefer: If you just need an approval, sure, here you go!

roschaefer (approved):

deployment/scripts/cluster.backend-bash.sh
roschaefer: You can do port-forwarding, SSH, logs and a lot more with `k9s`. I guess the features of `k9s` all cover what the bash scripts are trying to accomplish.

PR #7984, opened 2024-10-22
refactor(webapp): degroup vue packages in dependabot

mahula:
## 🍰 Pullrequest
In the dependabot configuration the grouping of all Vue related packages does not work for the update process.

### Issues
- relates #7951


Tirokk (approved):
Okidoki 🙏🏼

PR #8045, opened 2024-10-29
refactor(other): dependabot: further package update regroupings

mahula:
Motivation
------
In the dependabot configuration some more package update groupings do not work for the update process.


Tirokk (approved):
Supi

PR #8128, opened 2024-12-19
fix(frontend): hide invite button on mobile if disabled

ulfgebhardt:
## 🍰 Pullrequest

hide invite button on mobile if disabled

The invite button was displayed even when the functionality was dsiabled. This was caused by a missing check for the corresponding flag.

### Issues

- None

### Todo

- [X] None


webapp/components/HeaderMenu/HeaderMenu.vue
```diff
@@ -185,9 +185,11 @@
...
             style="text-align: center"
           >
-            <client-only>
-              <invite-button placement="top" />
-            </client-only>
+            <div v-if="inviteRegistration">
+              <client-only>
+                <invite-button placement="top" />
+              </client-only>
+            </div>
```
Tirokk, line 192: Unfortunately this makes a gap in the mobile header menu:

Bildschirmfoto 2025-01-04 um 13 13 20

So I suggest to remove the `if` property from the `div` tag to the superior `ds-flex-item` and remove the newly added `div` again.
This looks like this:

Bildschirmfoto 2025-01-04 um 13 22 47

Tirokk: Code is then:

```js
          <!-- invite button mobile -->
          <ds-flex-item
            v-if="inviteRegistration"
            :class="{ 'hide-mobile-menu': !toggleMobileMenu }"
            style="text-align: center"
          >
            <client-only>
              <invite-button placement="top" />
            </client-only>
          </ds-flex-item>
```

Tirokk (changes requested):
Cool that you cared @ulfgebhardt 

But there is a solution with a better design and layout.

Tirokk (approved):
I pushed my suggestion with your authorization.

Now we can merge I guess.

webapp/components/HeaderMenu/HeaderMenu.vue
Tirokk: @ulfgebhardt authorized me to push my suggestion.

PR #8255, opened 2025-03-04
fix(other): fix normalize e-mail gmail.com addresses on registration

### LINKED ISSUE #8252: 🐛 [Bug] Webapp – Registration has Error if we Login with `x.y@gmail.com` E-Mail
## 🐛 Bugreport

Registration has Error if we Login with `x.y@gmail.com` E-Mail

On registration with `wolle.huss@gmail.com` the edit field changes the text instantly to `wollehuss@gmail.com` (missing dot). Because this ends in registration of a wrong e-mail.

Image

### Todo

- [ ] fix change of `x.y@gmail.com` e-mail addresses by removing the dot in the first part

Tirokk:
## 🍰 Pullrequest

Fix normalize e-mail gmail.com addresses on registration.

### Issues

- fixes #8252 

### Todo

- [X] None

### DIFF
--- a/backend/src/schema/resolvers/helpers/normalizeEmail.spec.ts
+++ b/backend/src/schema/resolvers/helpers/normalizeEmail.spec.ts
@@ -1,0 +1,15 @@
+import normalizeEmail from './normalizeEmail'
+
+describe('normalizeEmail', () => {
+  it('converts all uppercase to lowercase', () => {
+    expect(normalizeEmail('Wolle.huSS@pJannto.com')).toBe('wolle.huss@pjannto.com')
+  })
+
+  it('removes not dots in gmail.com addresses', () => {
+    expect(normalizeEmail('wolle.huss@gmail.com')).toBe('wolle.huss@gmail.com')
+  })
+
+  it('converts not googlemail.com addresses to gmail.com', () => {
+    expect(normalizeEmail('wolle.huss@googlemail.com')).toBe('wolle.huss@googlemail.com')
+  })
+})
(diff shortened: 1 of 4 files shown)


Tirokk:
Backend test is failing:

`login › ask for a `token` › with a valid email/password combination › normalization › email address is a gmail address  › supplied email contains dots › normalizes email, issue #2329`

The old issue on Human Connection is: https://github.com/Human-Connection/Human-Connection/issues/2329

ulfgebhardt (commented):
I am unsure this is the right solution.

### Why?

I assume, we have databases which have normalized emails - hence the people will no longer be able to login as before.

### Options
1. Verify on all our installations gmail addresses are not normalized in the database
2. or normalize the emails as before and hook the normalizer in the login form.

### Further investigation

I check https://wir.social and the email from @Elweyn is a gmail address without dot (as shown in the admin view). As I believe @Elweyn normally writes it with a dot.
Maybe he can help to test this in conjunction with wir.social.

Executing `MATCH (e:EmailAddress) WHERE e.email ENDS WITH "@gmail.com" RETURN e;` on wir.social reveals 14 emails which all seem normalized.

### Different approach

After thinking about it and investigating further I think we could remove the normalize function completely in frontend and just do it in the backend. We don't care how the user writes his email in the frontend. We just make sure that in the backend its normalized. Here we can see some occurrences of the normalization.

### Good guess

I believe now that it is not a problem and a User with a gmail address can login normally, but he is confused where the dot goes.

1. check if we actually have a problem here (@Elweyn would have reported it if he could not login?!)
2. if there is a problem, ensure we fix it backend side
3. consider to remove the normalization in the frontend.

Lets talk about this.

PR #8256, opened 2025-03-06
feat(other): release script also update helmchart versions

ulfgebhardt:
## 🍰 Pullrequest

release script also update helmchart versions

To have congruence a release also updates the helmchart version. This utilizes sed to replace the coresponding string. The `sed` command is not present on windows and could prevent a windows user to use the release functionality out of the box.

The release functionality is also externalized into its own script `scripts/release.sh`.

### Issues

- None

### Todo

- [X] None


mahula (approved):
Nice.
That's way more cleaner and readable.

PR #8292, opened 2025-03-25
feat(backend): observe posts

Mogge:
## 🍰 Pullrequest

We want to have the possibility to observe posts:
 * a newly created post is always observed by the author
 * the first comment to a post makes the commenting user observing the post
 * provides the `toggleObservePost` mutation to switch observation on and off
 * unit tests that show the correct behavior, e.g. commenting a post that I do not observe for  a second time, does not alter the state of observation
 * adds `observedByMe` property to the Post type
 * provides a blueprint for the mutation in the webapp

### Issues
- relates #8280 

### Todo

- [X] None

### DIFF
--- a/backend/src/graphql/posts.ts
+++ b/backend/src/graphql/posts.ts
@@ -47,6 +47,7 @@ export const createPostMutation = () => {
           lat
         }
         observedByMe
+        observingUsersCount
       }
     }
   `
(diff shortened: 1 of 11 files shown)


backend/src/graphql/posts.ts
```diff
@@ -46,6 +46,7 @@ export const createPostMutation = () => {
           lng
           lat
         }
+        observedByMe
```
Bettelstab, line 49:
```suggestion
        isObservedByMe
```

Bettelstab (approved):
Works as expected

PR #8307, opened 2025-03-31
feat(backend): migrate observes relation between author and post

Mogge:
### Issues
- relates #8306 

### Todo

- [X] None

### DIFF
--- a/backend/src/db/migrations/20250331130323-author-observes-own-post.ts
+++ b/backend/src/db/migrations/20250331130323-author-observes-own-post.ts
@@ -1,0 +1,61 @@
+import { getDriver } from '../../db/neo4j'
+
+export const description = `
+All authors observe their posts.
+`
+
+export async function up(next) {
+  const driver = getDriver()
+  const session = driver.session()
+  const transaction = session.beginTransaction()
+
+  try {
+    // Implement your migration here.
+    await transaction.run(`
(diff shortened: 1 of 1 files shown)


Tirokk (approved):
Super @Mogge 

Works like expected 😍
I approve! 🚀💫

backend/src/db/migrations/20250331130323-author-observes-own-post.ts
```diff
@@ -0,0 +1,61 @@
...
+  const driver = getDriver()
+  const session = driver.session()
+  const transaction = session.beginTransaction()
+
+  try {
+    // Implement your migration here.
+    await transaction.run(`
+      MATCH (:User)-[obs:OBSERVES]->(p:Post)
+      DELETE obs
+      RETURN p
```
ulfgebhardt, line 47: Here you could also apply the down only to authors to match the migration, but meh thats not important

ulfgebhardt (approved):

Tirokk:
The seeding needs to be adjusted appropriate.
There we see that it might be better to use mutations in the seeding instead of `Factory.build()`.

ulfgebhardt (dismissed):
Good point @Tirokk 

I think this could be the appropriate PR to adjust the seeding aswell

PR #8308, opened 2025-03-31
feat(backend): migrate commenting users to observe commented posts

Mogge:
### Issues
- relates #8306 

### Todo

- [X] None


backend/src/db/migrations/20250331140313-commenter-observes-post.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export async function up(next) {
+  const driver = getDriver()
+  const session = driver.session()
+  const transaction = session.beginTransaction()
+
+  try {
+    // Implement your migration here.
+    await transaction.run(`
+      MATCH (commenter:User)-[:COMMENTS]->(post:Post)
```
Tirokk, line 15:
```suggestion
      MATCH (commenter:User)-[:WROTE]->(:Comment)-[:COMMENTS]->(post:Post)
```

This reaches your goal.
Your doesn't work.

Cypher to query in the database:

`MATCH (author:User)-[:WROTE]->(comment:Comment)-[:COMMENTS]->(post:Post) RETURN author, post;`

ulfgebhardt (changes requested):
I like this.
Compared to your other migration the down is more precise - please consider to adjust the other migration accordingly.

But we have factory seeded comments like those
```
await Factory.build(
  'comment',
  {
    id: 'c5',
  },
  {
    author: jennyRostock,
    postId: 'p3',
  },
),
```

Please investigate if the appropriate observe is set with those comments. Consider to move to mutated comments.

Tirokk (changes requested):
This needs to be fixed.

backend/src/db/migrations/20250331140313-commenter-observes-post.ts
Mogge: Thanks for finding this!

Tirokk (changes requested):
Please overtake my suggestions in case.

ulfgebhardt (commented):
I still get the following:

image

on `clean-migrate-scripts`:
```
yarn db:resetwithmigrations
yarn db:migrate init
yarn db:migrate up // exceutes 20250331130323-author-observes-own-post.ts
yarn db:seed
```

image
First post of jenny has 1 observer. Its not Jenny. Its Dewey

image

` MATCH (p:Post {id: 'p14'})-[r]-(n) RETURN p,r,n`
image

Why only Dewey? Why not Peter?

on `commenter-observes-post-migration`:
```
yarn db:migrate up
// executes
// 20231017141022-fix-event-dates.ts
// 20250331130323-author-observes-own-post.ts
// 20250331140313-commenter-observes-post.ts
```

image

image

---

Summary:
1. The migration works
2. Why is on master only one commenter an observer? Is There a mistake? Or related to a seed?
3. Why does the thing execute the last two migrations twice? `fix-event-dates`, `author-observes-own-post`

I would approve this if you and  @Tirokk think this is fine especially with the other PR of the seed. its just weird that there is one observer and not two.

Tirokk (dismissed):
Cool 😍

I approve 🚀💫

Tirokk:
There's the backend lint failing

Tirokk (approved):

PR #8311, opened 2025-04-01
refactor(backend): comment on observed post notification

Mogge:
## 🍰 Pullrequest

Instead of only notifying the author of the post, all users that observe the post are notified when the post is commented.

### Issues
- relates #8280 

### Todo

- [X] None


backend/src/middleware/notifications/notificationsMiddleware.spec.ts
```diff
@@ -238,7 +238,6 @@ describe('notifications', () => {
           })
 
           it('sends me no notification', async () => {
-            await notifiedUser.relateTo(commentAuthor, 'blocked')
```
Mogge, line 241: This is wrong!

ulfgebhardt (approved):
Works for me, but the seed does not work properly, as the observe is not correctly set initially

PR #8312, opened 2025-04-01
feat(webapp): track online status

ulfgebhardt:
## 🍰 Pullrequest

Regularly update online status of user.
This saves a timestamp on the user as `lastActiveAt` and a status `lastOnlineStatus` which can either be `online` or àway`.

The combination of the `lastActiveAt` timestamp and the `lastOnlineStatus` allows to determine if the user is reachable via notification or an email is the way to go.

image
In the screenshot you can see a call as `away` when the tab is switched and a call as `online` when the tab is activated again. The next entry is the regular ping every 30 seconds.

The `lastActiveAt` timestamp is set whenever the token of the user is seen querying the backend. A mechanic, which was already in place.

The Database looks like this:
image

### Issues

- contributes to https://github.com/Ocelot-Social-Community/Ocelot-Social/issues/8290

### Todo

- [x] remove console.log

### DIFF
--- a/backend/src/middleware/permissionsMiddleware.ts
+++ b/backend/src/middleware/permissionsMiddleware.ts
@@ -462,6 +462,7 @@ export default shield(
       switchUserRole: isAdmin,
       markTeaserAsViewed: allow,
       saveCategorySettings: isAuthenticated,
+      updateOnlineStatus: isAuthenticated,
       CreateRoom: isAuthenticated,
       CreateMessage: isAuthenticated,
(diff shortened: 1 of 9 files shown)


Mogge (commented):
I assume, that the online status should be provided to other users? Or is it just a information that holds the server in order how to handle notifications?

Mogge (changes requested):
I log in, my status is `online`. I close the tab, my status stays `online`

PR #8313, opened 2025-04-02
feat(backend): notify users when a followed user posted

Mogge:
### Issues

- None

### Todo

- [X] None


backend/src/middleware/notifications/notificationsMiddleware.ts
```diff
@@ -147,6 +149,43 @@ const postAuthorOfComment = async (commentId, { context }) => {
   }
 }
 
+const notifyFollowingUsers = async (postId, groupId, context) => {
+  if (groupId) return []
```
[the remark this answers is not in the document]
Mogge: I understand your desire, but this means, that we have to trigger another query to determine the group type. I really would like to avoid this query, which would be triggered every time a new post is published. We already have a lot of additional database operations. But if you insist ...

ulfgebhardt (commented):
👍🏻

backend/src/middleware/notifications/followed-users.spec.ts
```diff
@@ -0,0 +1,351 @@
...
+        mutation: createPostMutation,
+        variables: {
+          id: 'post',
+          title: 'This is the post',
+          content: 'This is the content of the post',
+        },
+      })
+    })
+
+    it('sends no notification to the post author', async () => {
```
Tirokk, line 154:
```suggestion
    it('sends NO notification to the post author', async () => {
```

backend/src/middleware/notifications/followed-users.spec.ts
Tirokk, line 1: Do we also want to ensure that no one outside the group is notified for hidden groups?

Tirokk (dismissed):

ulfgebhardt (commented):
Not a requirement, I understand why you did two queries, but using this mechanic allows you to reduce it to just one:

```
MATCH (u:User)
WHERE (u.email in $userEmails)
OPTIONAL MATCH (u)-[r:MEMBER_OF]-(t:Team)
WITH u, r, t WHERE t IS NULL OR NOT (t.id IN $excludedTeamIds)
RETURN r, u, t;
```
Source: https://stackoverflow.com/a/76380516

PR #8331, opened 2025-04-05
feat(backend): notification settings backend

ulfgebhardt:
## 🍰 Pullrequest

complex email notification settings

Query
```
query {currentUser{id emailNotificationSettings{name type value}}}
```
image

Update
```
mutation Update($update: [EmailNotificationSettingsInput])  {UpdateUser(id: "u4", emailNotificationSettings: $update){emailNotificationSettings{name type value}}}

{
  "update": [{"name": "groupMemberJoined", "value": false}]
}
```
image

### Issues

- requires https://github.com/Ocelot-Social-Community/Ocelot-Social/pull/8314 to finish notification tests

### Todo

- [x] Tests
- [x] Notification tests
- [ ] Frontend https://github.com/Ocelot-Social-Community/Ocelot-Social/pull/8320
- [ ] E2E


backend/src/schema/resolvers/users.ts
```diff
@@ -355,6 +381,42 @@ export default {
...
+          name: 'groupMemberRemoved',
+          type: 'group',
+          value: parent.emailNotificationsGroupMemberRemoved ?? true,
+        },
+        {
+          name: 'groupMemberRoleChanged',
+          type: 'group',
+          value: parent.emailNotificationsGroupMemberRoleChanged ?? true,
+        },
+      ]
```
Bettelstab, line 418:
```suggestion
      return [
  {
    type: 'post',
    settings: [
      {
        name: 'commentOnObservedPost',
        value: parent.emailNotificationsPostByFollowedUser ?? true,
      },
      {
        name: 'mention',
        value: parent.emailNotificationsPostInGroup ?? true,
      },
    ],
  },
  {
    type: 'group',
    settings: [
      {
        name: 'groupMemberJoined',
        value: parent.emailNotificationsGroupMemberJoined ?? true,
      },
      {
        name: 'groupMemberLeft',
        value: parent.emailNotificationsGroupMemberLeft ?? true,
      },
      {
        name: 'groupMemberRemoved',
        value: parent.emailNotificationsGroupMemberRemoved ?? true,
      },
      {
        name: 'groupMemberRoleChanged',
        value: parent.emailNotificationsGroupMemberRoleChanged ?? true,
      },
    ],
  },
]
```

webapp/graphql/User.js
```diff
@@ -390,7 +393,11 @@ export const currentUserQuery = gql`
       locale
       allowEmbedIframes
       showShoutsPublicly
-      sendNotificationEmails
+      emailNotificationSettings {
+        name
+        type
+        value
```
Bettelstab, line 399:
```suggestion
        type
        settings {
          name
          value
        }
```

Bettelstab (dismissed):
Feel free to use suggestions. Otherwise I can do the necessary transform in the frontend.

backend/src/schema/resolvers/users.ts
```diff
@@ -150,6 +150,31 @@ export default {
...
         params.termsAndConditionsAgreedAt = new Date().toISOString()
       }
+
+      const {
+        emailNotificationSettings,
+      }: { emailNotificationSettings: { name: string; value: boolean }[] | undefined } = params
+      delete params.emailNotificationSettings
+      if (emailNotificationSettings) {
+        emailNotificationSettings.forEach((setting) => {
+          const allowedSettingNames = [
```
Mogge, line 160: Why not defining this in the GraphQL Schema?

ulfgebhardt: How would you define such a komplex logic in the gql?

Mogge: I refer to the allowed names. Something like this:
```
enum NotificationReason {
  mentioned_in_post
  mentioned_in_comment
  commented_on_post
  user_joined_group
  user_left_group
  changed_group_member_role
  removed_user_from_group
}
```

Mogge: This would reduce the code of validation at this point

Mogge (commented):
I really do not understand why the type of the setting is in any case relevant for the database and therefore the whole backend. I understand, that having this structure in the client improves the UX a lot, but I cannot see the advantage of having it in the server as well. For me, those are simple, flat boolean values stored under an given name in the database

Mogge (commented):
I would have done something like:
```graphql
input _EmailSettings {
  commentOnObservedPost: Boolean
  mention: Boolean
  chatMessage: Boolean
  groupMemberJoined: Boolean
  groupMemberLeft: Boolean
  groupMemberRemoved: Boolean
  groupMemberRoleChanged: Boolean
}
```
This input may be passed to update user. Is then flattened in the resolver and done.

ulfgebhardt:
> Member

I both cases we have to adjust the frontend each time we introduce a new setting.

With the way its currently implemented we only need to add locales and never have to touch the frontend again. The backend defines how its structured in the frontend (the type) and we are agnostic to what kind of settings are available. 99% of magic happens in the backend now.

ulfgebhardt:
merged in https://github.com/Ocelot-Social-Community/Ocelot-Social/pull/8320

PR #8336, opened 2025-04-06
refactor(workflow): add scopes deps and deps-dev to pr linting workflow

mahula:
## 🍰 Pullrequest
Pull requests about dependencies should contain the relation to `deps`/`deps-dev` in their title.
Both scopes were not covered in the pr linting workflow and led to pr title linting errors.

### Issues
- relates #8263

### Todo

- [X] None


ulfgebhardt (commented):
The dependabot is excluded from this action - see https://github.com/Ocelot-Social-Community/Ocelot-Social/blob/master/.github/workflows/test.lint_pr.yml#L18

The bug is not the scopes but why this workflow is triggered

PR #8320, opened 2025-04-04
feat(webapp): notification settings frontend

Bettelstab:
## 🍰 Pullrequest
Implements detailed settings for email notifications.

### Issues
- relates #8280 

### Todo
- [ ] Fix initialising (dumb component receiving props vs created hook)
- [ ] Tests
- ...

PR #8347, opened 2025-04-08
fix(webapp): better settings ux

Bettelstab:
## 🍰 Pullrequest
This PR aims to improve the settings UX. It does two things:
- On mobile, scroll to the settings content when a menu entry is clicked
- Avoid that, on desktop and slow internet connection, the settings content appears first below the menu, sometimes staying there for quite a while:

Screenshot 2025-04-08 at 20 03 33

### Todo
- [ ] Tests. I propose to wait for #8320 with `vue-testing-library`. It could be a bit difficult to test the mixin, let's see. If not, maybe just add a test to `notifications.spec.js`?


ulfgebhardt (approved):
works as expected

PR #8354, opened 2025-04-10
refactor(backend): user graphql

ulfgebhardt:
## 🍰 Pullrequest

refactor user graphql

- remove isLoggedIn query
- currentUser query only for authenticated, currenUser always returns a User
- currentUser query implementation uses neo4jgraphql with id parameter
- remove custom email field from user
- fix frontend since currentUser now returns all categories

### Issues

- None

### Todo

- [X] None


Mogge (approved):
Nice and clean refactor

PR #8359, opened 2025-04-11
chore(backend): add e-mail setting for our new 'mailserver' to our backend .env.template

Tirokk:
### Issues

- None

### Todo

- [X] None


backend/.env.template
```diff
@@ -18,6 +18,12 @@ SMTP_SECURE="false" # true for 465, false for other ports
 SMTP_DKIM_DOMAINNAME=
 SMTP_DKIM_KEYSELECTOR=
 SMTP_DKIM_PRIVATKEY=
+# E-Mail settings for our 'docker compose up mailserver'
+# SMTP_HOST=localhost
```
ulfgebhardt, line 22:
```suggestion
# SMTP_HOST=mailserver
```

ulfgebhardt (changes requested):
Those settings didn't work on my maschine - the reason why it took me an hour to test some PRs.

Did any of you test the new software? Is it some weird setting on my maschine?

@mahula

backend/.env.template
Tirokk: I guess your suggestion doesn't work.

I tried it and was right. It doesn't. `localhost` is correct.

ulfgebhardt (approved):

PR #8363, opened 2025-04-11
fix(backend): block user

ulfgebhardt:
## 🍰 Pullrequest

block user

blocking a user resulted in an error. Blocking was not possible.

image

- made it possible
- simplified query into one

### Issues

- None

### For another PR

- [ ] There seems to be no block test
- [ ] Unify muted & blocked
- [ ] Muting has no effect


Mogge (approved):
No unit tests for this. Follow up issue?

PR #8364, opened 2025-04-11
fix(backend): block/mute chat

ulfgebhardt:
## 🍰 Pullrequest

block/mute chat
- No chat notification pubsub/email when user has blocked/muted the sender
- Does not show a notification bubble for blocked/muted users
- Also moves the pubsub into the notification middleware

Note: Chatting is still possible, but user is not informed when sender is blocked/muted

### Issues

- None

### Todo

- [x] Also when muted?
- [x] Test for proper behaviour


mahula (approved):
Good

PR #8392, opened 2025-04-15
fix(backend): mentioned users in posts and comments of groups

### LINKED ISSUE #8360: 🐛 [Bug]mention notifications to users that are no group members in posts in groups
## 🐛 Bugreport

It must be assured, that users that are mentioned in a group post  of a closed or hidden group they are no members do not receive notifications for this post.

Mogge:
## 🍰 Pullrequest

Ensures that no mentioned in post or comment notifications are sent to users that do not have access to the corresponding post or comment.

### Issues
- fixes #8360 

### Todo

- [X] None


backend/src/middleware/notifications/mentions-in-grops.spec.ts
```diff
@@ -0,0 +1,845 @@
...
+  mutate = createTestClientResult.mutate
+})
+
+afterAll(async () => {
+  await cleanDatabase()
+  driver.close()
+})
+
+describe('mentions in groups', () => {
+  beforeAll(async () => {
```
ulfgebhardt, line 118: I believe this test setup is flawed. With a `beforeAll` we cannot exchange the order of the testes within this file, something I do quite often when looking at old Testst.

As of my personal developer experience its hell to work with this kind of tests, because you do not know the state of the database and you have to read big file to understand what is happening, understand all previous tests.

The argument for this technique is flawed as well in my believe. It cannot be a problem to construct 10 Nodes and delete 10 Nodes from the database each test. Its not like we seed the database, its literally a hand full of nodes and relations each test. If this would really have a significant speed impact as we can read in a lot of comments, I am unsere we can serve any number of users with it.

I suggest a refactor to a `beforeEach`, which in fact should be all that has to be done. Addtnionally steps that the current test depends on have to be repeated.
This makes the tests atomic, exchangeable and understandable.

backend/src/middleware/notifications/mentions-in-grops.spec.ts
```diff
@@ -0,0 +1,845 @@
+import { createTestClient } from 'apollo-server-testing'
```
ulfgebhardt, line 1: I like the spit of the tests but I would wish for a prefix of the filename where the tests resides in. This is currently in a subfolder with one resolver, which makes this possible, but thats flawed in itself.

backend/src/middleware/notifications/mentions-in-grops.spec.ts
```diff
@@ -0,0 +1,845 @@
...
+              read: false,
+              reason: 'mentioned_in_post',
+            },
+          ],
+        },
+        errors: undefined,
+      })
+    })
+
+    it('sends 5 emails, three mentions and 2 post in group', () => {
```
ulfgebhardt, line 373: context is completley missing

Should send only 1 email? I am unsure

backend/src/middleware/notifications/mentions-in-grops.spec.ts
```diff
@@ -0,0 +1,845 @@
...
+              read: false,
+              reason: 'mentioned_in_post',
+            },
+          ],
+        },
+        errors: undefined,
+      })
+    })
+
+    it('sends 2 emails, one mention and one post in group', () => {
```
ulfgebhardt, line 468: it should only send 1 email, right? 1 user is notified if i am correct?

Again the missing context makes it hard to understand.

backend/src/middleware/notifications/mentions-in-grops.spec.ts
```diff
@@ -0,0 +1,845 @@
...
+  describe('post in hidden group', () => {
+    beforeAll(async () => {
+      jest.clearAllMocks()
+      authenticatedUser = await noMember.toJson()
+      await markAllAsRead()
+      authenticatedUser = await pendingMember.toJson()
+      await markAllAsRead()
+      authenticatedUser = await groupMember.toJson()
+      await markAllAsRead()
+      authenticatedUser = await postAuthor.toJson()
```
ulfgebhardt, line 482: with an atomic test setup all this will be gone

backend/src/middleware/notifications/mentions-in-grops.spec.ts
```diff
@@ -0,0 +1,845 @@
...
+              reason: 'post_in_group',
+            },
+            {
+              from: {
+                __typename: 'Post',
+                id: 'public-post',
+              },
+              read: false,
+              reason: 'mentioned_in_post',
+            },
```
ulfgebhardt, line 366: should be only 1

backend/src/middleware/notifications/mentions-in-grops.spec.ts
```diff
@@ -0,0 +1,845 @@
...
+          ],
```
ulfgebhardt, line 333: should send only one

backend/src/middleware/notifications/mentions-in-grops.spec.ts
```diff
@@ -0,0 +1,845 @@
...
+        },
+        errors: undefined,
+      })
+    })
+
+    it('sends a notification to the pending member', async () => {
```
ulfgebhardt, line 305: Do we have pending members in public groups? I thought you can just join

backend/src/middleware/notifications/notificationsMiddleware.ts
```diff
@@ -358,9 +359,12 @@ const notifyUsersOfMention = async (label, id, idsOfUsers, reason, context) => {
       mentionedCypher = `
       MATCH (postAuthor: User)-[:WROTE]->(post: Post)<-[:COMMENTS]-(comment: Comment { id: $id })<-[:WROTE]-(commenter: User)
       MATCH (user: User)
-      WHERE user.id in $idsOfUsers
-      AND NOT (user)-[:BLOCKED]-(commenter)
-      AND NOT (user)-[:BLOCKED]-(postAuthor)
+        WHERE user.id in $idsOfUsers
+        AND NOT (user)-[:BLOCKED]-(commenter)
```
ulfgebhardt, line 363: we also have to include muted for Notifications. Muted is included in Blocked. Muted silences all notifications.

backend/src/middleware/notifications/notificationsMiddleware.ts
```diff
@@ -346,9 +346,10 @@ const notifyUsersOfMention = async (label, id, idsOfUsers, reason, context) => {
     case 'mentioned_in_post': {
       mentionedCypher = `
         MATCH (post: Post { id: $id })<-[:WROTE]-(author: User)
-        MATCH (user: User)
-        WHERE user.id in $idsOfUsers
-        AND NOT (user)-[:BLOCKED]-(author)
+        MATCH (user: User) WHERE user.id in $idsOfUsers AND NOT (user)-[:BLOCKED]-(author)
```
ulfgebhardt, line 349: same here, see below, mentioned/blocked

ulfgebhardt (commented):
I think part of my review is probably out of scope for this particular issue. I just don't understand why you write tests to ensure behaviour we explicitly don't want.

I really would like a different test setup; I know there is a lot of comments about the slowness of this method - let us discuss this please. I do not understand how we came to that conclusion.

backend/src/middleware/notifications/mentions-in-grops.spec.ts
Mogge: Nope. At least the group member always has a notification, that the role in the group changed

backend/src/middleware/notifications/mentions-in-grops.spec.ts
Mogge: Statement from discord:
```
Was denkst du dazu @moriz ?
moriz — gestern um 18:29 Uhr
Wenn ein Autor, dem ich folge einen Post in eine meiner Gruppen schreibt und mich darin erwähnt, dann bekomme ich drei Benachrichtigungen, soll aber nur eine Email bekommen. Ist das so richtig verstanden?
Wolle (Hamburg) — gestern um 18:31 Uhr
So ist es richtig verstand.

Drei Benachrichtigungen, aber nur eine E-Mail. @Ulf Gebhardt 
````
So what do we want?

backend/src/middleware/notifications/mentions-in-grops.spec.ts
Mogge: Same

backend/src/middleware/notifications/notificationsMiddleware.ts
Mogge: So then, what is the difference between muted and blocked?

backend/src/middleware/notifications/mentions-in-grops.spec.ts
Mogge: Same here

backend/src/middleware/notifications/mentions-in-grops.spec.ts
Mogge: I do not understand. What do you want, can you give an example?

ulfgebhardt: id like the tests to be named `notificationsMiddleware.*.spec.ts` respective `notificationsMiddleware.mentions-in-grops.spec.ts` to ensure that the files can be associated with the actual file to be tested even when moved into another folder with more files.

ulfgebhardt: btw `grops` should be `groups`

backend/src/middleware/notifications/mentions-in-grops.spec.ts
ulfgebhardt: i think its bad practice to write tests for behaviour we actually want to prevent, but for the sake of the separation of different PR parts, I think it is acceptable.

backend/src/middleware/notifications/notificationsMiddleware.ts
ulfgebhardt: blocked should also hide all content of an user, muted should only hide notifications. But blocked also mutes the notifications.

Don't expect this to work properly, but we should ensure correct behaviour on notifications.

Mogge: I would like to do this in another PR, as testing is a lot of code and this issue gets out of scope

ulfgebhardt (dismissed):

ulfgebhardt (approved):

PR #8394, opened 2025-04-16
feat(backend): no mention notification for non group members new

Mogge:
### Issues

- None

### Todo

- [X] None


Mogge:
merged into #8392

PR #8401, opened 2025-04-16
feat(webapp): badges admin settings

Bettelstab:
## 🍰 Pullrequest
Adds a link to badges settings in the user table, where admins can set the available badges.

### Issues
- relates #8291 
- 
### Todo
- [ ] Show icons instead of text
- [ ] Styling
- [ ] Tests


ulfgebhardt (approved):
Tested and approved.

ulfgebhardt (commented):
This happens if there is no badges in the database.
image

We might want to introduce a config settings if badges are available

Issue #8388, opened 2025-04-14
🔧 [Refactor] Refine Notifications


ulfgebhardt:
## 🔧 Refactor ticket

Refine Notifications

After testing the notifications on https://wir.social we noticed the following problems:

- [x] Invite to Group - Setting missing
- [ ] Only 1 email not up to 3 (group post, mentioned, following the user) (here a priority should be defined e.g. mentioned > group > follow)
- [ ] The notification emails should always feature the reason for the notification and the user causing it. The subject should already reflect a portion of that information. In order for the user to decide if he wants to open the email.
- [x] EMail notifications only when the user is offline - see chat notifications
- [ ] Ensure notifications are not sent when the use is either muted or blocked
- [x] Filter mentions/followUser in private groups to notify only when the notified has access to the content
- [ ] Investigate what the reason could be that users are addressed with the wrong name in some cases (happend on wir.social)
- [ ] (optional) use [pug multi-language emails](https://www.npmjs.com/package/email-templates)

![Image](https://github.com/user-attachments/assets/39887e64-6211-4a64-8078-533c2d0ceb2e)

Mogge:
Please see my comment on discord to the second point.

Mogge:
Please see my comment on discord to the first point

PR #8403, opened 2025-04-16
feat(backend): do not notify blocked or muted users

Mogge:
## 🍰 Pullrequest

BLOCKED works in both directions. Meaning a blocked user does not receive notifications from me.
MUTED only works in one direction. Meaning a muted user may receive notifications from me.

### Issues
- relates #8388

### Todo

- [X] None

### DIFF
--- a/backend/src/middleware/notifications/notificationsMiddleware.spec.ts
+++ b/backend/src/middleware/notifications/notificationsMiddleware.spec.ts
@@ -294,6 +294,25 @@ describe('notifications', () => {
               ).resolves.toEqual(expected)
             })
           })
+
+          describe('if I have muted the comment author', () => {
+            it('sends me no notification', async () => {
+              await notifiedUser.relateTo(commentAuthor, 'muted')
+              await createCommentOnPostAction()
+              const expected = expect.objectContaining({
+                data: { notifications: [] },
+              })
+
+              await expect(
+                query({
+                  query: notificationQuery,
+                  variables: {
+                    read: false,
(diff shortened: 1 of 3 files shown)


backend/src/middleware/notifications/notificationsMiddleware.spec.ts
```diff
@@ -736,6 +797,72 @@ describe('notifications', () => {
             expect(pubsubSpy).toHaveBeenCalledTimes(1)
           })
         })
+
+        describe('but the author of the post muted me', () => {
```
ulfgebhardt, line 801:
```suggestion
        describe('but the author of the comment muted me', () => {
```

ulfgebhardt (commented):
think just a copy typo, else its good to go

backend/src/middleware/notifications/notificationsMiddleware.spec.ts
Mogge: No, this is the author of the post:
```
        AND NOT (user)-[:BLOCKED]-(commenter)
        AND NOT (user)-[:BLOCKED]-(postAuthor)
        AND NOT (user)-[:MUTED]->(commenter)
        AND NOT (user)-[:MUTED]->(postAuthor)
```

Mogge: The test shows, that is has no effect to the notification

ulfgebhardt (approved):

PR #8419, opened 2025-04-21
fix(backend): fix notification emails with different name

ulfgebhardt:
## 🍰 Pullrequest

fix different name notifications

We had emails sent with incorrect names.
This PR combines the query for the email with the user the notification is sent to, since the notification in database was correct. The underlying problem is the unstable order in which the database can return values. The results of the two queries were matched by id since it was assumed that they always return the same order of elements.

### Issues

- releates https://github.com/Ocelot-Social-Community/Ocelot-Social/issues/8388

### Todo

- [X] None


Mogge (changes requested):
As you have in the code the case, that there is no email available, it would be good to test it as well, e.g. create one user with no attached email and expect that no email is sent. One small improvement of the tests would be to change the email addresses as well, e.g. `email: 'test4@example.org -> email: 'third.follower@example.org''`

Thank you for fixing this. My assumption was the same, that the way how the email was derived was the problem.

ulfgebhardt:
> As you have in the code the case, that there is no email available, it would be good to test it as well, e.g. create one user with no attached email and expect that no email is sent. One small improvement of the tests would be to change the email addresses as well, e.g. `email: 'test4@example.org -> email: 'third.follower@example.org''`
> 
> Thank you for fixing this. My assumption was the same, that the way how the email was derived was the problem.

those emails were defined by you

Mogge (approved):

PR #8421, opened 2025-04-21
refactor(backend): allow to set selected badge-slot to null

ulfgebhardt:
## 🍰 Pullrequest

allow to set selected badge-slot to null

Free a specific badge slot by setting it to null

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/schema/resolvers/users.ts
+++ b/backend/src/schema/resolvers/users.ts
@@ -414,7 +414,7 @@ export default {
           `
         const queryNull = `
             MATCH (user:User {id: $userId})
-            OPTIONAL MATCH (user)-[slotRelation:SELECTED{slot: $slot}]->(:Badge)
+            OPTIONAL MATCH (user)-[slotRelation:SELECTED {slot: $slot}]->(:Badge)
             DELETE slotRelation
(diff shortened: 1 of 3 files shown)


backend/src/schema/resolvers/users.ts
```diff
@@ -404,17 +404,25 @@ export default {
...
             DELETE badgeRelation, slotRelation
             MERGE (user)-[:SELECTED{slot: toInteger($slot)}]->(badge)
             RETURN user {.*}
-          `,
-          { userId, badgeId, slot },
-        )
+          `
+        const queryNull = `
+            MATCH (user:User {id: $userId})
+            OPTIONAL MATCH (user)-[slotRelation:SELECTED{slot: $slot}]->(:Badge)
```
Tirokk, line 417:
```suggestion
            OPTIONAL MATCH (user)-[slotRelation:SELECTED {slot: $slot}]->(:Badge)
```

Tirokk (dismissed):
Works well! 🥰

Just a little suggestion, because we always put a space there to have it better readable …

Bettelstab (approved):
works as expected

PR #8425, opened 2025-04-22
feat(backend): lint - detect unused typescript disables

ulfgebhardt:
## 🍰 Pullrequest

lint - detect unused typescript disables

### Issues

- None

### Todo

- [X] None


backend/src/schema/resolvers/helpers/databaseLogger.ts
```diff
@@ -1,21 +0,0 @@
-/* eslint-disable @typescript-eslint/no-unsafe-call */
```
roschaefer, line 1: We're never going to log database calls anymore?

roschaefer (approved):
very nice

backend/src/schema/resolvers/helpers/databaseLogger.ts
ulfgebhardt: We have zero logging - we need to change this. But his half baked debug capability utilizing an package not even in our `package.json` is more trouble then worth to keep it. Just my opinion.

roschaefer (dismissed):

roschaefer (approved):

PR #8434, opened 2025-04-24
refactor(backend): refactor context

ulfgebhardt:
## 🍰 Pullrequest

refactor context

first steps to improve the context handling

### Issues

- based on https://github.com/Ocelot-Social-Community/Ocelot-Social/pull/8409

### Todo

- [X] None


backend/src/schema/resolvers/users.spec.ts
```diff
@@ -108,25 +109,28 @@ const resetTrophyBadgesSelected = gql`
   }
 `
 
+const database = databaseContext()
```
Mogge, line 112:
```suggestion
const database = databaseContext()
const { neode } = database
```
So we do not need the following changes in the file

Mogge (dismissed):
Why are the calls for `contextUser` sometimes unsafe and sometimes not?

backend/src/schema/resolvers/users.spec.ts
ulfgebhardt: I do this deliberately since we want to get rid of neode. This way all calls are to `database.neode` which might be `database.query` or `database.User.create` in the future.

ulfgebhardt:
> Why are the calls for `contextUser` sometimes unsafe and sometimes not?

It is always `@typescript-eslint/no-unsafe-return` and `@typescript-eslint/require-await`. Sometimes the ignore is global, sometimes just on the particular line, depending on the test case.
While `@typescript-eslint/no-unsafe-return` can be fixed by defining the Database Types for user, the await cant be fixed, since the actual context requires a Promise and we return a resolved object.

Tirokk (approved):

PR #8437, opened 2025-04-25
fix(backend): revokeBadge also removes selection

ulfgebhardt:
## 🍰 Pullrequest

revokeBadge also removes selection

On current master the selection of a badge is not removed when the admin revokes the badge. This way the badge is still shown on the profile even tho the user doesn't own it anymore. Only when the user unselect the badge he cannot reassign it.

This PR fixes that and removes the selection when revoking a Badge as well.

### Issues

- None

### Todo

- [X] None


Bettelstab (commented):
works

Bettelstab (approved):

PR #8438, opened 2025-04-25
fix(webapp): fix admin badges settings

Bettelstab:
## 🍰 Pullrequest
- Fixes the proxying of icons: In #8401, I modified the nuxt config to proxy requests to /img to the backend, unaware of the existing vue filter solution. This PR reverts the config change and uses the filter instead.
- Shows message when there are no badges in the database which can be rewarded. To avoid that the message appears while loading, I only show the `BadgeSelection` after badges have been loaded.

relates #8291


ulfgebhardt (commented):
I do not understand what you do here.

Bettelstab:
> I do not understand what you do here.

I updated the description, hope now it makes more sense.

ulfgebhardt (approved):

PR #8450, opened 2025-04-27
feat(webapp): user teaser popover

### LINKED ISSUE #8428: 🚀 [Feature] User Teaser
## 🚀 Feature
When hovering (desktop) or clicking (mobile) the user name or avatar, there should appear a user teaser popup containing the following information:
- Distance to user
- Verification badge
- Trophy badges
- Do I follow the user?
- Other follow information
- A button to open the profile, possibly only on mobile

Bettelstab:
## 🍰 Pullrequest
Implements the user teaser popover.
Screenshot 2025-04-30 at 22 55 52

Screenshot 2025-04-30 at 22 53 48

In comparison to the old implementation in human-connection, the popover is only rendered when actually used.

### Issues
- fixes #8428 

### Todo
- [x] Prevent post from being opened on mobile, when clicking on user name in news feed
- [x] Tests
- [x] Go to user profile on Desktop
- [x] SSR
- [x] Fix notifications hover crash
- [x] Fix group members hover crash
- [x] No distance on profile, news feed, post page.
- [x] Try to fit in popover on smaller screens
- [x] Clickable spans should be buttons
- [x] Load missing data
- [x] Update tests


ulfgebhardt (changes requested):
It is not possible to go to the users profile from the teaser in desktop mode.

image

Clicking on any user part will open the popup; clicking again closes it. I can just navigate to the post.

The user profile is reachable via search only.

---

Hovering over a user field in notifications makes the app crash

image
image

If you are fast you can click the username on the notifications, nothing happens, nothing crashes if you leave fast enough before the popup opens

---

Another crash when hovering over
image

---

Switching to mobile requires a hard reload to have the Button show and vice versa

image

image

---

Serverside rendering of the user profile crashes the page

image

ulfgebhardt:
Suggestion:

Put the click handler for the user on a large field

image

I'd actually like to have click boxes like this
image

bt I beleive this could be alot more refactoring - but that would be ideal

ulfgebhardt (changes requested):
Hovering over a user in the notifications closes the popup and shows this empty one
image


https://github.com/user-attachments/assets/ecc6c90f-4425-4624-bbc4-018566e59b0d

ulfgebhardt (commented):
When the popup opens the scroll bar disappears. This is very irritating (certainly acceptable for now), but I would prefer the scroll bar to stay visible to not have a blinking sensation. It also moves the page when inserting the scrollbar, which causes flickering.

A solution could be not to gray out the surroundings, but popup in place, without modifying the rest of the page.

Artifacts

image
image

Scrolling no longer possible
image

ulfgebhardt (dismissed):

ulfgebhardt (dismissed):

ulfgebhardt (approved):

PR #8455, opened 2025-04-29
fix(webapp): fixed padding for mobile in basic layout

sebastian2357:
## 🍰 Pullrequest

the basic layout "blank" contained a hardcoded style attribute with a padding, which was a way too big for mobile.
This PR fixes the issue.

### Issues

- None

### Todo

- [X] Fix

### DIFF
--- a/webapp/layouts/blank.vue
+++ b/webapp/layouts/blank.vue
@@ -17,11 +17,11 @@ export default {
 </script>
 
 <style lang="scss">
-.ds-container > div {
+.layout-blank > .ds-container > div {
   padding: 5rem 2rem;
 }
 @media only screen and (max-width: 500px) {
-  .ds-container > div {
+  .layout-blank > .ds-container > div {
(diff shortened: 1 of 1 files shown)


Tirokk (changes requested):
Great that you care @sebastian2357 !

But I got this `default` layout.

Bildschirmfoto 2025-04-30 um 12 57 35

webapp/layouts/blank.vue
```diff
@@ -15,3 +15,14 @@ export default {
   mixins: [seo],
 }
 </script>
+
+<style lang="scss">
```
Tirokk, line 19: Maybe use

```vue
<style lang="scss" scoped>
```

instead of

```vue
<style lang="scss">
```

Tirokk: We find an other solution.

Tirokk (approved):
Cool, I approve

PR #8484, opened 2025-05-03
[WIP] refactor(backend): jest - global server and database connection

ulfgebhardt:
## 🍰 Pullrequest

jest - global server and database connection

This should speed up the testing by alot, since the server just has to be started once.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/jest.config.cjs
+++ b/backend/jest.config.cjs
@@ -23,5 +23,7 @@ module.exports = {
   },
   testMatch: ['**/src/**/?(*.)+(spec|test).ts?(x)'],
   setupFilesAfterEnv: ['<rootDir>/test/setup.ts'],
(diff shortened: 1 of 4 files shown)


roschaefer (changes requested):
danger

@ulfgebhardt have you done a performance benchmark?

Because I question the speed benefits from this PR.

I would assume that the apollo server setup costs nothing. What makes test execution slow is cleaning and writing the database. Of course you can omit that. But then tests may interfere with each other.

The cleaner solution to achieve fast test execution is *not to use a database* but to mock it.

Complexity wise I would strongly suggest to go to the exact opposite direction of this PR. Do not pollute `global` and avoid shared mutual state because that is the root of all evil.

PR #8486, opened 2025-05-03
refactor(backend): types for context + `slug`

roschaefer:
Also these changes saw merge conflicts in #8463 so let's get them merged already.


backend/src/server.ts
```diff
@@ -102,3 +102,4 @@ const createServer = (options?) => {
 }
 
 export default createServer
+export type Context = Awaited<ReturnType<ReturnType<typeof getContext>>>
```
roschaefer, line 105: Why does `getContext` have to return two nested functions?

ulfgebhardt: Because it needs to have context, and requires to be a function to evaluate during runtime.
If you have a better way to inject dependencies, feel free to implement it

ulfgebhardt (approved):

PR #8517, opened 2025-05-09
feat(webapp): redirect on registration for invite links

Mogge:
## 🍰 Pullrequest
When the user is logged in and gets an invite code, then:
 - it checks the invite code
 - it redirects to index if it is no group invite
 - or it redeems the invite code to the group and redirects to the group if it is public, else to index

### Issues

- None

### Todo

- [X] None


webapp/graphql/inviteCodes.ts
```diff
@@ -0,0 +1,19 @@
+import gql from 'graphql-tag'
```
ulfgebhardt: remove

webapp/pages/registration.spec.js
```diff
@@ -330,6 +349,92 @@ describe('Registration', () => {
       expect(redirect).toHaveBeenCalledWith('/')
     })
 
+    describe('already logged in', () => {
+      beforeEach(async () => {
+        asyncData = true
+        isLoggedIn = true
+      })
+
+      describe('route contains public invite code', () => {
```
ulfgebhardt, line 358:
```suggestion
      describe('route contains personal invite code', () => {
```

ulfgebhardt (approved):

PR #8519, opened 2025-05-09
fix(backend): copy of constants in dockerfile

Mogge:
### Issues

- None

### Todo

- [X] None


ulfgebhardt (approved):
I approve because I must

PR #8544, opened 2025-05-11
fix(backend): package.json migrate scripts

roschaefer:
I see merge conflicts again, so let's merge this independent change
already. I think this is probably due to a package update.

Error fixed is:
```
$ migrate --compiler 'ts:./src/db/compiler.ts' --migrations-dir ./src/db/migrations --store ./src/db/migrate/store.ts
error: unknown option `--compiler'
error Command failed with exit code 1.
```

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -21,9 +21,9 @@
     "db:data:badges": "ts-node --require tsconfig-paths/register src/db/badges.ts",
     "db:data:branding": "ts-node --require tsconfig-paths/register src/db/data-branding.ts",
     "db:data:categories": "ts-node --require tsconfig-paths/register src/db/categories.ts",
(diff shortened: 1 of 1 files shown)


ulfgebhardt (changes requested):
We cannot merge this - there is more then 1 command to call with `db:migrate` - you cannot hardcode it.
I don't see how the order of commands has any impact on this.
The current master runs on all our machines and on production.

roschaefer:
> We cannot merge this - there is more then 1 command to call with `db:migrate` - you cannot hardcode it. I don't see how the order of commands has any impact on this. The current master runs on all our machines and on production.

Ah wait, the script is not supposed to be called without a command line argument?

ulfgebhardt:
correct you call it with `up` or `init`

PR #8545, opened 2025-05-11
feat(backend): migrate to s3

roschaefer:
## 🍰 Pullrequest
This will migrate our assets to an objectstorage via S3.

Before this PR is rolled out, the S3 credentials need to be configured in the respective infrastructure repository. The migration is implemented in a backend migration, i.e. I expect the `initContainer` to take a little longer but I hope then it's going to be fine. If any errors occcur, the migration should be repeatable, since the disk volume is still there.

### Issues
The backend having direct access on disk.

### Todo
- [ ] Configure backend environment variables in every infrastructure repo


backend/src/db/migrate/store.ts
```diff
@@ -36,8 +36,8 @@ class Store {
 
       // You have to wait for the schema to install, else the constraints will not be present.
       // This is a type error of the library
-      // eslint-disable-next-line @typescript-eslint/await-thenable
-      await getNeode().schema.install()
+
+      getNeode().schema.install()
```
ulfgebhardt, line 40: you must wait it - it doesnt work properly if you don't - I tested it - see comment above

backend/src/db/migrations/20250502230521-migrate-to-s3.ts
```diff
@@ -0,0 +1,109 @@
...
+import CONFIG from '@config/index'
+import { getDriver } from '@db/neo4j'
+
+export const description =
+  'Upload all image files to a S3 compatible object storage in order to reduce load on our backend.'
+
+export async function up(_next) {
+  if (CONFIG.NODE_ENV === 'test') {
+    // Let's skip this migration for simplicity.
+    // There is nothing to migrate in test environment and setting up the S3 services seems overkill.
```
ulfgebhardt, line 19: i agree that it is not necessary for test environment to migrate stuff, but since the s3 will be the only option for image handling we need to set it up as well. The E2E tests for example test for correct behaviour if I am correct.

backend/src/db/migrations/20250502230521-migrate-to-s3.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export async function down(_next) {
+  const driver = getDriver()
+  const session = driver.session()
+  const transaction = session.beginTransaction()
+
+  try {
+    // Implement your migration here.
+    await transaction.run(``)
```
ulfgebhardt, line 97: i suggest to just throw here and forget about it

backend/src/graphql/resolvers/images/imagesLocal.ts
```diff
@@ -0,0 +1,134 @@
+/* eslint-disable @typescript-eslint/require-await */
```
ulfgebhardt, line 1: will  we still support local images? You remove some parts of the corresponding code - the migration throws an error if you do not have an s3 configured.

backend/src/graphql/resolvers/images/sanitizeRelationshipTypes.ts
```diff
@@ -0,0 +1,7 @@
+export const sanitizeRelationshipType = (relationshipType: string) => {
+  // Cypher query language does not allow to parameterize relationship types
+  // See: https://github.com/neo4j/neo4j/issues/340
+  if (!['HERO_IMAGE', 'AVATAR_IMAGE'].includes(relationshipType)) {
```
ulfgebhardt, line 4: probably not your problem, but there are solutions to this problem
- For create with apoc: https://github.com/neo4j/neo4j/issues/340#issuecomment-793219411
- For query with normal cypher: https://github.com/neo4j/neo4j/issues/340#issuecomment-356265721

backend/src/db/migrate/store.ts
roschaefer: That's so strange. So we must *not* `await` here? Is that right?

backend/src/db/migrations/20250502230521-migrate-to-s3.ts
roschaefer: OK! I can for sure setup a test S3 service - it might slow down CI a bit.

backend/src/graphql/resolvers/images/imagesLocal.ts
roschaefer: I was hesitating to remove the code before the migration is done. Maybe I will remove it straight away when I refactor (e.g. pass s3Client via context) so I don't have to maintain it.

backend/src/graphql/resolvers/images/sanitizeRelationshipTypes.ts
roschaefer: Yes, this code was here before. Maybe I can add this as a separate PR.

backend/src/db/migrate/store.ts
ulfgebhardt: The types are broken. It is a Promise thats returned. This library is just a nightmare.

backend/src/graphql/resolvers/images/imagesLocal.ts
ulfgebhardt: Ok, jsut for my understanding - its fine to keep the code for a moment for reference

backend/src/db/migrations/20250502230521-migrate-to-s3.ts
ulfgebhardt: You will find an appropriate solution - just have a look at https://github.com/Ocelot-Social-Community/Ocelot-Social/blob/master/cypress/e2e/UserProfile.Avatar.feature

backend/src/graphql/resolvers/images/sanitizeRelationshipTypes.ts
roschaefer: > Apologies for the intrusion. I found this PR via the open issues linked above, and I wanted to let you know that Neo4j now fully supports dynamic labels and types as of versions 2025.03 and 5.26.4 LTS, released in March of this year.
> 
> You can read more about them [here](https://neo4j.com/blog/developer/cypher-dynamism/), or read the documentation [here](https://neo4j.com/docs/cypher-manual/current/clauses/match/#dynamic-match).

That's great to hear! Unfortunately we're stuck with ecosystem and dependencies in nodejs. It will take some effort before we'll be able to upgrade to v5.

Biggest pain: https://github.com/neo4j-graphql/neo4j-graphql-js

But thank you for letting us know!

backend/src/graphql/resolvers/images/imagesS3.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+
+    it('deletes `Image` node', async () => {
+      await expect(neode.all('Image')).resolves.toHaveLength(1)
+      await deleteImage(user, 'AVATAR_IMAGE', { deleteCallback })
+      await expect(neode.all('Image')).resolves.toHaveLength(0)
+    })
+
+    it('calls deleteCallback', async () => {
+      const u = await Factory.build('user')
+      user = await u.toJson()
```
ulfgebhardt, line 82: Why construct it again, you have a `beforeEach` in line 60?

backend/src/graphql/resolvers/images/imagesS3.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+        await expect(
+          mergeImage(post, 'HERO_IMAGE', imageInput, { uploadCallback, deleteCallback }),
+        ).resolves.toMatchObject({
+          url: expect.stringMatching(
+            new RegExp(`^http://your-objectstorage.com/bucket/${uuid}-foo-bar-avatar.jpg`),
+          ),
+        })
+      })
+
+      describe('but given a `S3_PUBLIC_GATEWAY` configuration', () => {
```
ulfgebhardt, line 209:
```suggestion
      describe('given a `S3_PUBLIC_GATEWAY` configuration', () => {
```

ulfgebhardt:
```suggestion
      describe('with a `S3_PUBLIC_GATEWAY` configuration', () => {
```

deployment/helm/charts/ocelot-social/templates/backend/stateful-set.yaml
```diff
@@ -25,6 +25,9 @@ spec:
               name: {{ .Release.Name }}-backend-env
           - secretRef:
               name: {{ .Release.Name }}-backend-secret-env
+          volumeMounts:
+            - mountPath: /app/public/uploads
+              name: uploads
```
ulfgebhardt, line 30: Just for my understanding - this is to be able to migrate?!

ulfgebhardt (commented):
Some small changes and questions.

Further more:
- If a file does not exist, we can no longer migrate the service. How do we want to handle this?
image

deployment/helm/charts/ocelot-social/templates/backend/stateful-set.yaml
roschaefer: Yes, that's right. Otherwise the `initContainer` cannot upload the files to `S3`. Before, it wouldn't need to have access because the migrations wouldn't touch the uploads folder.

Later, we should remove this volume access from both container definitions.

backend/src/graphql/resolvers/images/imagesS3.spec.ts
roschaefer: Sure, there is no test case before that explicitly says what happens when you don't have it set.

roschaefer:
> Some small changes and questions.
> 
> Further more:
> 
>     * If a file does not exist, we can no longer migrate the service. How do we want to handle this?
>       image

We had a quick chat and I would suggest to let the migration crash in that case. It would mean we have an image in the database but no corresponding file on disk. We should probably remove that `Image` node manually from the database.

PR #8550, opened 2025-05-12
fix(backend): fix statistics and introduce new values

ulfgebhardt:
## 🍰 Pullrequest

fix statistics and introduce new values
- more values
- correct invites count
- simplify frontend logic and made it easily extensible

image
image

### Issues

- None

### Todo

- [X] None


Tirokk (changes requested):
Have some fixes suggested.

The sum over the Members of all groups could be interesting as well.

Tirokk (approved):

PR #8555, opened 2025-05-14
fix(webapp): fix popover flickering

Bettelstab:
## 🍰 Pullrequest
- Add boundary to v-popover to avoid random flickering when hovering a user teaser
 - Preload user data to avoid resizing (and sometimes repositioning) of popover after data has been loaded


ulfgebhardt (approved):
I can no longer reproduce the flickering (was possible on master for Jenny's followers)

PR #8558, opened 2025-05-15
feat(backend): image transformations

roschaefer:
### Issues

- None

### Todo

- [X] None


Bettelstab:
@roschaefer What would be your plan regarding the chat component? Forking?

webapp/components/_new/features/Admin/Badges/BadgesSection.spec.js
```diff
@@ -5,14 +5,14 @@ const localVue = global.localVue
 
 const badge1 = {
   id: 'badge1',
-  icon: 'icon1',
```
roschaefer: We rely on the leading `'/'` here.

webapp/components/_new/features/Admin/Badges/BadgesSection.vue
```diff
@@ -8,7 +8,7 @@
         @click="toggleBadge(badge)"
         :class="{ badge, inactive: !badge.isActive }"
       >
-        <img :src="badge.icon | proxyApiUrl" :alt="badge.description" />
+        <img :src="'/api' + badge.icon" :alt="badge.description" />
```
ulfgebhardt: I prefer a filter here. This way we could also omit the leading slash on the `icon` attribute. (See File above in review)

Why I prefer a filter? If we change the endpoint of the api e.g. from `domain/api` to `api.domain` we have a trillion occurrences, we have to update.

rather do:
```suggestion
        <img :src="badge.icon | apiUrl " :alt="badge.description" />
```

This way you can also enforce the `/` if not present. Preferable with https://developer.mozilla.org/en-US/docs/Web/API/URL/URL .

webapp/components/_new/features/Admin/Badges/BadgesSection.spec.js
ulfgebhardt: I don't like this solution - have a look at https://github.com/Ocelot-Social-Community/Ocelot-Social/pull/8558#discussion_r2195816598

webapp/components/_new/features/Admin/Badges/BadgesSection.vue
ulfgebhardt: Could also be `backenUrl` as filter of `fromBackend` - whatever suits best

roschaefer: Filters have been removed in Vue V3: https://v3-migration.vuejs.org/breaking-changes/filters

backend/src/graphql/resolvers/images.ts
```diff
@@ -1,9 +1,84 @@
...
+
+const changeDomain: (opts: { transformations: UrlResolver[] }) => UrlResolver =
+  ({ transformations }) =>
+  ({ url }, _args, context) => {
+    const { config } = context
+    const { S3_PUBLIC_GATEWAY, AWS_ENDPOINT } = config
+    if (!S3_PUBLIC_GATEWAY) {
+      return url
+    }
+    const originalUrl = new URL(url, AWS_ENDPOINT) // some S3 object storages return invalid URLs as location, so put the AWS_ENDPOINT as `base`
```
ulfgebhardt: are you sure this works?
We have urls like this: 
> fsn1.your-objectstorage.com/BUCKET/original/uploads/UUID-bildschirmfoto-2025-05-10-um-001013.png

They were missing the `https://` which resulted the browser to resolve this to:
> https://my.tld/fsn1.your-objectstorage.com/BUCKET/original/uploads/UUID-bildschirmfoto-2025-05-10-um-001013.png

backend/src/graphql/resolvers/messages.ts
```diff
@@ -13,6 +13,8 @@ import { CHAT_MESSAGE_ADDED } from '@constants/subscriptions'
 import { attachments } from './attachments/attachments'
 import Resolver from './helpers/Resolver'
 
+import type { File } from './attachments/attachments'
```
ulfgebhardt, line 16: unused?

docker-compose.override.yml
```diff
@@ -39,7 +40,8 @@ services:
       - AWS_ENDPOINT=http:/minio:9000
       - AWS_REGION=local
       - AWS_BUCKET=ocelot
-      - S3_PUBLIC_GATEWAY=http:/localhost:9000
+      - S3_PUBLIC_GATEWAY=http:/localhost:8000
```
ulfgebhardt: Rename for clarity as this only applies to images now?

```suggestion
      - S3_IMAGE_GATEWAY=http:/localhost:8000
```

webapp/components/BadgeSelection.vue
```diff
@@ -7,7 +7,7 @@
       @click="handleBadgeClick(badge, index)"
     >
       <div class="badge-icon">
-        <img :src="badge.icon | proxyApiUrl" :alt="badge.id" />
+        <img :src="'/api' + badge.icon" :alt="badge.id" />
```
ulfgebhardt: I still believe `/api` should be a constant, but i wont block the merge for this.

roschaefer: Oh, I can do this, no problemo.

docker-compose.override.yml
roschaefer: We have the same brain. I already renamed it while working locally.

backend/src/graphql/resolvers/images.ts
roschaefer: That's extremely annoying, yes. 

On a side note, I'm still confused how this can even happen:
* In docker environment, the returned `Location` is `http://minio:9000` as if the client library defines the host from its point of view
* In production we get these errors, because of a weird behaviour of Hetzner's S3 storage as if the server defines the host from its point of view

The way I understand it is that docker sets the `hostname` for the container and `minio` just picks it up. Ie. the server decides the host of the `Location` field.

backend/src/graphql/resolvers/messages.ts
roschaefer: Nope. There is a global `File` type. If we don't import our own custom type with that name, typescript believes we want to use the built-in type.

deployment/helm/charts/ocelot-social/templates/ingress.yaml
```diff
@@ -62,4 +62,34 @@ spec:
...
+
+---
+apiVersion: traefik.io/v1alpha1
+kind: Middleware
+metadata:
+  name: {{ .Release.Name }}-stripprefix
+spec:
+  stripPrefix:
+    prefixes:
+      - /imagor
```
roschaefer, line 75: We're tunneling to the backend through the webapp - which is imperformant (needs to go through NodeJS runtie). We could do the same thing for `backend` service here.

backend/src/uploads/s3Service.ts
```diff
@@ -40,13 +40,7 @@ export const s3Service = (config: S3Config, prefix: string) => {
       location = `https://${location}`
     }
 
-    if (!S3_PUBLIC_GATEWAY) {
-      return location
-    }
-
-    const publicLocation = new URL(S3_PUBLIC_GATEWAY)
-    publicLocation.pathname = new URL(location).pathname
-    return publicLocation.href
```
roschaefer, line 49: Let's not write that to the database but replace the domain host on-the-fly during a request.

roschaefer: We're lucky this environment value didn't make it on production yet.

deployment/helm/helmfile/secrets/ocelot.yaml
```diff
@@ -23,15 +23,25 @@ secrets:
             NEO4J_USERNAME: null
             NEO4J_PASSWORD: null
             REDIS_PASSWORD: null
-            AWS_ACCESS_KEY_ID: ENC[AES256_GCM,data:iiN5ueqyo60VHb9e2bnhc19iGTg=,iv:zawYpKrFafgsu1+YRet1hzZf1G3a6BIlZgsh7xNADaE=,tag:rTsmm8cqei34b6cT6vn08w==,type:str]
-            AWS_SECRET_ACCESS_KEY: ENC[AES256_GCM,data:Zl4LRXdDh/6Q8F9RVp+3L7NXGZ0F2cgFMKPhl/TVeuD5Bhy68W5ekg==,iv:AmPoinGISrSOZdoBKdeFFXfr2hwOK4nWMnniz8K5qgU=,tag:K8Q7M7e+6G9T0Oh3Sp4OzA==,type:str]
-            AWS_ENDPOINT: ENC[AES256_GCM,data:/waEqUgcOmldZ+peFTNVsDQf2KrpWY8ZZMt1nT5117SkbY4=,iv:n+Kvidjb/TM4bQYKqTaFxt8GkHo02PuxEGpzgOcywr4=,tag:lrGPgCWWy3GMIcTv75IYTg==,type:str]
-            AWS_REGION: ENC[AES256_GCM,data:kBPpHZ8zw4PMpg==,iv:R+QZe303do37Hd/97NpS1pt9VaBE/gqZDY2/qlIvvps=,tag:0WduW8wfJXtBqlh4qfRGNA==,type:str]
-            AWS_BUCKET: ENC[AES256_GCM,data:0fAspN/PoRVPlSbz+qDBRUOieeC4,iv:JGJ/LyLpMymN0tpZmW6DjPT3xqXzK/KhYQsy9sgPd60=,tag:Y6PBs0916JkHRHSe7hqSMA==,type:str]
+            IMAGOR_SECRET: null
```
roschaefer, line 26: Don't forget to merge https://github.com/Ocelot-Social-Community/stage.ocelot.social/pull/21

docker-compose.override.yml
```diff
@@ -36,10 +39,12 @@ services:
       - SMTP_HOST=mailserver
       - AWS_ACCESS_KEY_ID=minio
       - AWS_SECRET_ACCESS_KEY=12341234
-      - AWS_ENDPOINT=http:/minio:9000
+      - AWS_ENDPOINT=http:/localhost:9000
```
roschaefer, line 42: The backend is the man-in-the-middle here.

backend/src/index.ts
```diff
@@ -14,3 +15,20 @@ httpServer.listen({ port: url.port }, () => {
...
+  const target = new URL(CONFIG.PROXY_S3)
+  const proxy = createProxy(target)
+  const forwardedPort = target.port // target port and forwarded port must be the same
+  proxy.listen(forwardedPort, () => {
+    /* eslint-disable-next-line no-console */
+    console.log(`Simple HTTP proxy listening on port ${forwardedPort}`)
+    /* eslint-disable-next-line no-console */
+    console.log(`Proxying requests to ${target}`)
+  })
+}
```
ulfgebhardt, line 34: Whats hindering us in configuring `http://localhost:9000` as endpoint instead of `http://minio:9000`?

backend/test/helpers.ts
```diff
@@ -16,6 +16,7 @@ export const TEST_CONFIG = {
   PRODUCTION_DB_CLEAN_ALLOW: false,
   DISABLED_MIDDLEWARES: [],
   SEND_MAIL: false,
+  PROXY_S3: 'http://minio:9000',
```
ulfgebhardt, line 19: why not `http://localhsot:9000` and not have the proxy code?

webapp/components/BadgeSelection.vue
```diff
@@ -17,6 +17,7 @@
 </template>
 
 <script>
+import { backendPath } from '~/helpers/backendPath'
```
ulfgebhardt, line 20: use mixin?

webapp/components/Badges.vue
```diff
@@ -8,12 +8,13 @@
...
       @click="handleBadgeClick(index)"
     >
-      <img :title="badge.description" :src="badge.icon | proxyApiUrl" class="hc-badge" />
+      <img :title="badge.description" :src="backendPath(badge.icon)" class="hc-badge" />
     </component>
   </div>
 </template>
 
 <script>
+import { backendPath } from '~/helpers/backendPath'
```
ulfgebhardt, line 17: use mixin?

webapp/components/_new/features/Admin/Badges/BadgesSection.vue
```diff
@@ -18,6 +18,7 @@
 </template>
 
 <script>
+import { backendPath } from '~/helpers/backendPath'
```
ulfgebhardt, line 21: use mixin?

webapp/helpers/backendPath.js
```diff
@@ -0,0 +1 @@
+export const backendPath = (url) => (url.startsWith('/') ? '/api' + url : '/api/' + url)
```
ulfgebhardt, line 1: i blieve this should be a mixin as this is the right mechanic to supply common methods to different components in vue.

docker-compose.override.yml
ulfgebhardt: I understand, but Why?

deployment/helm/charts/ocelot-social/templates/ingress.yaml
ulfgebhardt: Can you elaborate on this? The frontend runs on the client most of the time - or is the proxying happening on the server part of nuxt?

webapp/components/BadgeSelection.vue
ulfgebhardt: Thank you - please make it a mixin

backend/src/index.ts
roschaefer: We do use `http://localhost:9000` as the `AWS_ENDPOINT`! The backend container will upload to itself (`localhost`).

There is an explanation in the comment above. We need to make sure that the `AWS_ENDPOINT` is a domain that is both reachable from the backend container and the host. We cannot use a different domain name, as we would get a 404 for the images in the browser on the host machine.

Instead of changing the domain name in the application layer in the backend when uploading vs. when accessing through the GraphQL I decided to solve this problem on the network level by forwarding port 9000 on the backend (including `localhost:9000`) to `minio:9000`. Thus we can use the same name on the host machine. I hope that makes sense.

backend/test/helpers.ts
roschaefer: Without the proxy, the backend cannot upload to `localhost:9000`.

roschaefer:
> It works as expected:
> 
> This image 366kb: image
> 
> is requested as 190kb: image image
> 
> There seems to be something wrong with the fronted source set, but that shall not be the problem of this PR I believe. The source set contains the w320 size, but the sizes do not contain this. image
> 
> The fronted seems to request always the w1024 in my browser, even when simulating a phone.
> 
> I would like to merge this, but have two minor things I'd like to clarify first:
> 
>     1. Why use a proxy for local development when you can simply define http://localhost:9000 in the config?
> 
>     2. Can we make the backendApi prefixer a mixin and move it in the proper folder? Why would that not be applicable?
> 
> 
> Besides this I'd like to merge this! ❤️ Thank you @roschaefer this is great!

I've experienced the same behaviour on my machine as well. I expected smaller sizes to be loaded when on a mobile phone. Maybe our frontend devs @Bettelstab or @sebastian2357 know why?

webapp/helpers/backendPath.js
roschaefer: I can refactor of course. But mind this:
> In Vue 2, mixins were the primary mechanism for creating reusable chunks of component logic. While mixins continue to be supported in Vue 3, [Composable functions using Composition API](https://vuejs.org/guide/reusability/composables) is now the preferred approach for code reuse between components.

https://vuejs.org/api/options-composition#mixins

ulfgebhardt (approved):

PR #8559, opened 2025-05-16
fix(webapp): notifications - UI Improvements

sebastian2357:
Notifications:
- restructured broken layout
- joined several columns for mobile view
- moved button from footer to header
- set alternating colors for the table rows

UserTeaser:
- added injectedText
- added injectedDate
- fixed padding

### Issues

- None

### Todo

- [x] Notifications
 - [X] Layout
 - [x] Jumping Menu
 - [x] Read Flag does not work reliably
 - [x] NotificationMenu should use same component as notification page
 - [x] Popup get closed when user hovers scrollbar
 - [x] Popup background layer remains when popup disappeared by viewport switch


ulfgebhardt (commented):
image
broken layout

Tirokk (approved):
Very cool @sebastian2357 😍

Look really nice and I approve it! 🚀🚀💫💫

PS: That this menu partly and the filter menu if it's in the mobile header sticks out of the view and is only partly visible was already in the master and we are on the way to fix it on a different PR.

PR #8574, opened 2025-05-18
fix(webapp): improve login & invite mechanics

ulfgebhardt:
## 🍰 Pullrequest

improve login & invite mechanics

We have trouble with ~ 3% of user not able to click next when entering a valid invite code or nonce.
This PR simplifies some things to improve the register sliders. Since we currently cannot reproduce the problem this is just a guess.
- Removes reference to `$nextTick` functionality
- Reevaluate Invite-Codes and Nonces even when entering the same code again
- Hide Inviting user when code becomes invalid again

image
image
image
image
image

image

### Issues

- None

### Todo

- [X] None


webapp/components/Registration/RegistrationSlideInvite.vue
```diff
@@ -124,18 +122,11 @@ export default {
...
-        this.sliderData.sliders[this.sliderIndex].data.request.variables &&
-        this.sliderData.sliders[this.sliderIndex].data.request.variables.code === variables.code
-      )
-    },
     async handleSubmitVerify() {
       const { inviteCode } = this.sliderData.collectedInputData
       const variables = { code: inviteCode }
 
-      if (!this.isVariablesRequested(variables) && !this.dbRequestInProgress) {
+      if (!this.dbRequestInProgress) {
```
Bettelstab, line 129: Hmm. I think the purpose of `isVariablesRequested()` was to check if the same request was made before. Now it could happen that a request to check if a code is valid is not made, because _another_ code is being checked.

Bettelstab (changes requested):
Not sure about the removal of `isVariablesRequested`. Could be simplified for sure (e.g. just checking `this.sliderData.sliders[this.sliderIndex].data.request.variables.code === code`. Or disabling input during dbRequests - that might be the easiest.

No glue why the $nextTick functionality was used. Hope it doesn't matter, looks not like it would be needed. However, I tested the live version in MiBrowser and I have little hope that this will help. First, the invite code is not even read from the URL. And when manually entered, no reaction.
I could not find out how to debug in that thing, if even possible...

Bettelstab (approved):

PR #8575, opened 2025-05-18
KI "optimierter" test-e2e.yml

Hendrik-cpu:


.github/workflows/test-e2e.yml
```diff
@@ -1,179 +1,70 @@
...
+        job: [1, 2, 3, 4, 5, 6, 7, 8]
+    env:
+      MATRIX_SIZE: 8
     steps:
-      - name: Checkout code
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.1.7
+      - uses: actions/checkout@v4
 
-      - name: Build docker image
+      - name: Build and save docker images
```
Tirokk, line 26: maybe do this step before the matrix

.github/workflows/test-e2e.yml
```diff
@@ -1,179 +1,70 @@
...
-
-  prepare_cypress:
-    name: Fullstack | prepare cypress
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.1.7
-
-      - name: Copy env files
+          docker build --target community -t ocelotsocialnetwork/neo4j-community:test neo4j/
```
Tirokk, line 28: this part is no longer parallel - docker images are build sequentially

.github/workflows/test-e2e.yml
```diff
@@ -1,179 +1,70 @@
-name: ocelot.social end-to-end test CI
-
-on: push
+name: ocelot.social test-e2e ⇌⋇⟴
+
+on:
+  push:
+    paths:
```
Tirokk, line 5: this looks like a funny new syntax, what does it do?

docker-compose.test.yml
```diff
@@ -12,7 +12,7 @@ services:
 
   backend:
     # name the image so that it cannot be found in a DockerHub repository, otherwise it will not be built locally from the 'dockerfile' but pulled from there
-    image: ghcr.io/ocelot-social-community/ocelot-social/backend:test
+    image: ocelotsocialnetwork/backend:test
```
Tirokk, line 15: for testing it is ok to change images, but we need this to be set to the github image repository

Tirokk (commented):
Thank you for this - keep going - there is some interesting suggestions

PR #8592, opened 2025-05-25
fix(webapp): fix websocket - downgrade graphql

ulfgebhardt:
## 🍰 Pullrequest

fix websocket - downgrade graphql

The websocket now connects again and the webapp receives server pushes.

### Issues

- None

### Todo

- [X] None


Tirokk (approved):
Very good.

Thanks that you cared for! 🚀💫

PR #8593, opened 2025-05-25
fix(webapp): update user, write all values, query verification Badge

ulfgebhardt:
## 🍰 Pullrequest

update user, write all values, query verification Badge

Required for auto verification on reformer

### Issues

- None

### Todo

- [X] None


Tirokk (approved):
Seems good to me

PR #8598, opened 2025-05-26
feat(backend): pin more than one post

Mogge:
## 🍰 Pullrequest
 * new `env` variable in backend: `MAX_PINNED_POSTS` defaults to 1
 * if `MAX_PINNED_POSTS === 0` no posts can be pinned
 * if `MAX_PINNED_POSTS === 1` nothing changes
 * if `MAX_PINNED_POSTS > 1` post are not automatically unpinned, only max number can be pinned

### Issues

- None

### Todo

- [X] None


webapp/store/pinnedPosts.spec.js
```diff
@@ -0,0 +1,111 @@
+import { state as initialState, mutations, getters, actions } from './pinnedPosts'
+import { postsPinnedCountsQuery } from '~/graphql/PostQuery'
+
+describe('pinned post store', () => {
+  describe('initial state', () => {
+    it('sets all values to 0', () => {
+      expect(initialState()).toEqual({
+        maxPinnedPosts: 0,
```
Tirokk, line 8: Should this be `maxPinnedPosts: 1`? 🤔

Tirokk (changes requested):
Hey @Mogge ,
nicely done!

Some Questions and suggestions:

- Even if `MAX_PINNED_POSTS=0` I should be able to unpin an already pinned post. Or?
  - Because we can lower the possible pins to zero if already pins where made before.
- If all possible posts are pinned the post context menu item to pin a post is invisible.
  - Is this the right solution? Because the admin might be irritated if the item is missing and don't know why pinning is impossible.
    - We could have a modal dialogue as alternative which tells the admin what to do to be able to pin a new post. A dialogue might be done later.

Tirokk (approved):
Cool thing

PR #8551, opened 2025-05-12
refactor(webapp): store for categories

Mogge:
### Issues
- fixes #8543

### Todo

- [X] None

PR #8603, opened 2025-05-28
refactor: put CONFIG into context

roschaefer:
This is a side quest of #8558. The motivation is to be able to do dependency injection in the tests without overwriting global data. I saw the first merge conflict from #8551 and voila: It seems @Mogge could have used this already.


backend/src/middleware/notifications/notificationsMiddleware.emails.spec.ts
```diff
@@ -12,6 +12,7 @@ import Factory, { cleanDatabase } from '@db/factories'
 import { createGroupMutation } from '@graphql/queries/createGroupMutation'
 import { joinGroupMutation } from '@graphql/queries/joinGroupMutation'
 import CONFIG from '@src/config'
+import { TEST_CONFIG } from '@src/config/test-mock'
 import createServer, { getContext } from '@src/server'
 
 CONFIG.CATEGORIES_ACTIVE = false
```
Mogge, line 18: TEST_CONFIG?

backend/src/middleware/notifications/notificationsMiddleware.followed-users.spec.ts
```diff
@@ -10,6 +10,7 @@ import databaseContext from '@context/database'
 import Factory, { cleanDatabase } from '@db/factories'
 import { createGroupMutation } from '@graphql/queries/createGroupMutation'
 import CONFIG from '@src/config'
+import { TEST_CONFIG } from '@src/config/test-mock'
 import createServer, { getContext } from '@src/server'
 
 CONFIG.CATEGORIES_ACTIVE = false
```
Mogge, line 16: TEST_CONFIG?

backend/src/middleware/notifications/notificationsMiddleware.mentions-in-groups.spec.ts
```diff
@@ -13,6 +13,7 @@ import { changeGroupMemberRoleMutation } from '@graphql/queries/changeGroupMembe
 import { createGroupMutation } from '@graphql/queries/createGroupMutation'
 import { joinGroupMutation } from '@graphql/queries/joinGroupMutation'
 import CONFIG from '@src/config'
+import { TEST_CONFIG } from '@src/config/test-mock'
 import createServer, { getContext } from '@src/server'
 
 CONFIG.CATEGORIES_ACTIVE = false
```
Mogge, line 19: TEST_CONFIG

backend/src/middleware/notifications/notificationsMiddleware.observing-posts.spec.ts
```diff
@@ -10,6 +10,7 @@ import gql from 'graphql-tag'
 import CONFIG from '@config/index'
 import databaseContext from '@context/database'
 import Factory, { cleanDatabase } from '@db/factories'
+import { TEST_CONFIG } from '@src/config/test-mock'
 import createServer, { getContext } from '@src/server'
```
Mogge, line 15: same

backend/src/middleware/notifications/notificationsMiddleware.online-status.spec.ts
```diff
@@ -8,6 +8,7 @@ import gql from 'graphql-tag'
 import databaseContext from '@context/database'
 import Factory, { cleanDatabase } from '@db/factories'
 import CONFIG from '@src/config'
+import { TEST_CONFIG } from '@src/config/test-mock'
 import createServer, { getContext } from '@src/server'
 
 CONFIG.CATEGORIES_ACTIVE = false
```
Mogge, line 14: same

backend/src/middleware/notifications/notificationsMiddleware.posts-in-groups.spec.ts
```diff
@@ -13,6 +13,7 @@ import { changeGroupMemberRoleMutation } from '@graphql/queries/changeGroupMembe
 import { createGroupMutation } from '@graphql/queries/createGroupMutation'
 import { joinGroupMutation } from '@graphql/queries/joinGroupMutation'
 import CONFIG from '@src/config'
+import { TEST_CONFIG } from '@src/config/test-mock'
 import createServer, { getContext } from '@src/server'
 
 CONFIG.CATEGORIES_ACTIVE = false
```
Mogge, line 19: same

Mogge (changes requested):

backend/src/middleware/notifications/notificationsMiddleware.emails.spec.ts
roschaefer: You mean you would prefer to have the file named '@src/config/test-config'?

backend/src/middleware/categories.spec.ts
```diff
@@ -57,8 +46,20 @@ const categoriesQuery = gql`
 
 describe('categroeis middleware', () => {
   describe('categories are active', () => {
-    beforeEach(() => {
-      CONFIG.CATEGORIES_ACTIVE = true
+    beforeAll(() => {
```
ulfgebhardt, line 49: we have this block outside normally

backend/src/middleware/categories.spec.ts
```diff
@@ -77,8 +78,20 @@ describe('categroeis middleware', () => {
   })
 
   describe('categories are not active', () => {
-    beforeEach(() => {
-      CONFIG.CATEGORIES_ACTIVE = false
+    beforeAll(() => {
+      const authenticatedUser = null
```
ulfgebhardt, line 82: we have this block outside normally

backend/src/middleware/categories.ts
```diff
@@ -1,9 +1,19 @@
-/* eslint-disable @typescript-eslint/no-unsafe-return */
-/* eslint-disable @typescript-eslint/no-unsafe-call */
-import CONFIG from '@src/config'
+import type { Context } from '@src/server'
 
-const checkCategoriesActive = (resolve, root, args, context, resolveInfo) => {
-  if (CONFIG.CATEGORIES_ACTIVE) {
+type Resolver = (
```
ulfgebhardt, line 3: this must be defined somewhere - we should not define it ourselfs

backend/src/middleware/notifications/notificationsMiddleware.emails.spec.ts
```diff
@@ -12,6 +12,7 @@ import Factory, { cleanDatabase } from '@db/factories'
 import { createGroupMutation } from '@graphql/queries/createGroupMutation'
 import { joinGroupMutation } from '@graphql/queries/joinGroupMutation'
 import CONFIG from '@src/config'
```
ulfgebhardt, line 14: ???

backend/src/middleware/notifications/notificationsMiddleware.emails.spec.ts
ulfgebhardt, line 18: ???

backend/src/middleware/notifications/notificationsMiddleware.followed-users.spec.ts
```diff
@@ -10,6 +10,7 @@ import databaseContext from '@context/database'
 import Factory, { cleanDatabase } from '@db/factories'
 import { createGroupMutation } from '@graphql/queries/createGroupMutation'
 import CONFIG from '@src/config'
```
ulfgebhardt, line 12: ???

backend/src/middleware/notifications/notificationsMiddleware.followed-users.spec.ts
ulfgebhardt, line 16: ???

backend/src/server.ts
```diff
@@ -33,11 +33,13 @@ export const getContext =
       database = serverDatabase,
       pubsub = serverPubsub,
       user = databaseUser,
+      config = CONFIG,
     }: {
       database?: ReturnType<typeof databaseContext>
       pubsub?: ReturnType<typeof pubsubContext>
       user?: (any) => Promise<any>
-    } = { database: serverDatabase, pubsub: serverPubsub, user: databaseUser },
+      config: typeof CONFIG
```
ulfgebhardt, line 41: i prefer this to be an optional parameter.

As I see now there is a default value - maybe sync how this behaves

ulfgebhardt (changes requested):
I support the endeavour; putting the config in the context makes a lot of sense, as it is a dependency injected into the resolvers.

I do not like:
- That it is not an optional parameter (for tests). With the current implementation you must provide it, even tho you do not need it for testing. Which is contrary to the rest of the context, which currently is only then explicitly created in the tests when needed
- You cannot easily manipulate the config anymore in tests, which make them much more complex, as you need to construct the complete context again to just change a config value and test for different behaviour.

I suggest:
- Make the context magical - apply the mock when running as `NODE_ENV=test`, use the config in all other cases
- Make the mock a function you can override as it is done with the `authenticatedUser`. This way you can modify the context during a test run without reconstructing the context. Maybe not very puristic, but would ensure better readability for tests, shortens code and is way more flexible. For the production case you can evaluate the config once, store it in a variable and return its value in a function whenever requested.

Furthermore - part of this PR should be to remove all `import CONFIG from '@config/index`, or am I missing something? The whole mechanic is not used except for `backend/src/middleware/categories.ts`. I am almost certain there is more Resolvers referencing the config - they are just untested.

backend/src/server.ts
ulfgebhardt: I suggest
```suggestion
      config?: () => typeof CONFIG
```

roschaefer:
> I support the endeavour; putting the config in the context makes a lot of sense, as it is a dependency injected into the resolvers.
> 
> I do not like:
> 
>     * That it is not an optional parameter (for tests). With the current implementation you must provide it, even tho you do not need it for testing. Which is contrary to the rest of the context, which currently is only then explicitly created in the tests when needed
> 
>     * You cannot easily manipulate the config anymore in tests, which make them much more complex, as you need to construct the complete context again to just change a config value and test for different behaviour.
> 
> 
> I suggest:
> 
>     * Make the context magical - apply the mock when running as `NODE_ENV=test`, use the config in all other cases
> 
>     * Make the mock a function you can override as it is done with the `authenticatedUser`. This way you can modify the context during a test run without reconstructing the context. Maybe not very puristic, but would ensure better readability for tests, shortens code and is way more flexible. For the production case you can evaluate the config once, store it in a variable and return its value in a function whenever requested.
> 
> 
> Furthermore - part of this PR should be to remove all `import CONFIG from '@config/index`, or am I missing something? The whole mechanic is not used except for `backend/src/middleware/categories.ts`. I am almost certain there is more Resolvers referencing the config - they are just untested.

Cool, this was just a quick cherry-pick from another PR of mine. If this is an endeavour that is supported by the team I will create a helper function to setup an apollo server with test parameters that can be overwritten where necessary.

backend/src/server.ts
roschaefer: I wouldn't do that as it changes the type of `Context` which we reference in our resolvers. We would have to check for nullish all the time.

backend/src/middleware/categories.ts
roschaefer: There are ways to correctly set types for GraphQL resolvers based on your schema. But I found it a bit out of scope for this PR.

backend/src/server.ts
ulfgebhardt: I  agree on the `?` part

backend/src/context/fetch.ts
```diff
@@ -0,0 +1,16 @@
+import request from 'request'
+
+export const fetch = (url: string) => {
```
ulfgebhardt: i do not understand why you inject fetch into the context as well? Thats a lib function and should be free to use, has zero side effect, is not configured? Why inject?

backend/src/graphql/resolvers/groups.spec.ts
```diff
@@ -230,7 +226,11 @@ const seedComplexScenarioAndClearAuthentication = async () => {
 }
 
 beforeAll(async () => {
-  await cleanDatabase()
+  const apolloSetup = createApolloTestSetup({ contextUser, config: { CATEGORIES_ACTIVE: true } })
```
ulfgebhardt: Its still a static value and not a fucntion. I would like somethink liks this

```suggestion
  let config = { ...TEST_CONFIG, CATEGORIES_ACTIVE: true } 
  const contextConfig = () =>  config
  const apolloSetup = createApolloTestSetup({ contextUser, config: contextConfig })
```

This way you can override the config at a later point just like this:
```
  let config = { ...TEST_CONFIG, CATEGORIES_ACTIVE: false }
```
We construct the context only once.

This would be analog to the `authenticatedUser` as seen in all tests, which proves to be a very effective way to switch users: 

```
authenticatedUser = await admin.toJson()
somecommand()
authenticatedUser = await usertoJson()
someothercommand()
```
Very easy, very readable. I always optimize for readability first and especially when it concerns tests. They don't need to be fast, just maintainable.

Any reason why not to do this?

backend/src/graphql/resolvers/locations.ts
```diff
@@ -1,10 +1,12 @@
 /* eslint-disable @typescript-eslint/no-unsafe-return */
-/* eslint-disable @typescript-eslint/no-unsafe-call */
+
```
ulfgebhardt:
```suggestion
```

backend/src/graphql/resolvers/users/location.spec.ts
```diff
@@ -1,16 +1,30 @@
...
+let database: ApolloTestSetup['database']
+let server: ApolloTestSetup['server']
+
+const fetchMock: Context['fetch'] = (url) => {
+  const response = mapboxResponses[url] // eslint-disable-line security/detect-object-injection
+  if (!response) {
+    throw new Error(`Missing response for url: ${url}`)
+  }
+  return response // eslint-disable-line @typescript-eslint/no-unsafe-return
+}
```
ulfgebhardt: Why context fetch - i really don't get this. It has no side effects. By that logic we must mock `console.log`, `parseInt` and all other functionality?!

backend/src/middleware/hashtags/hashtagsMiddleware.spec.ts
```diff
@@ -1,21 +1,21 @@
 /* eslint-disable @typescript-eslint/no-unsafe-call */
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
-/* eslint-disable @typescript-eslint/no-unsafe-argument */
+
```
ulfgebhardt:
```suggestion
```

backend/src/middleware/notifications/notificationsMiddleware.emails.spec.ts
```diff
@@ -1,27 +1,27 @@
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/no-unsafe-call */
-/* eslint-disable @typescript-eslint/require-await */
+
```
ulfgebhardt:
```suggestion
```

backend/src/middleware/notifications/notificationsMiddleware.mentions-in-groups.spec.ts
```diff
@@ -1,28 +1,28 @@
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/no-unsafe-call */
-/* eslint-disable @typescript-eslint/require-await */
+
```
ulfgebhardt:
```suggestion
```

backend/src/middleware/notifications/notificationsMiddleware.posts-in-groups.spec.ts
```diff
@@ -1,28 +1,28 @@
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
-/* eslint-disable @typescript-eslint/require-await */
+
```
ulfgebhardt:
```suggestion
```

ulfgebhardt (commented):
I like this, but
- why the fetch in the context - I don't think this adds value?! Its used like once? If you do this for the mock, I am sure you can mock the library
- Changing the config in the tests is still to complex - I like a solution like for  the user. See some comment.

backend/src/middleware/categories.ts
ulfgebhardt: leave it at `any`? You can ignore it in the file and as soon as we properly type our apollo properly it shows up as no longer needed?!
It seems forced.

backend/src/graphql/resolvers/users/location.spec.ts
roschaefer: Oh, I found out that our CI is sending API requests on every build against Mapbox. That's a common bad practice. It slows down CI and also makes the build depend on the availability of an external API.

ulfgebhardt: I agree - but making the context cancer is not the best solution ;)
Mock the library in the test - I am 100% on board - not a context thing - maybe even a different PR.

roschaefer: What I don't like is that recording a request is cumbersome. I implemented the most naive approach with a few hard-coded api responses, expecting that we don't have many. But it's already 9 different urls we use somewhere in our tests.

There are helper libraries for this e.g [pollyjs](https://github.com/Netflix/pollyjs).

ulfgebhardt: It has nothing to do with this PR and I believe it should not be part of the context to mock it easier.

backend/src/context/fetch.ts
roschaefer: By the way, we don't use nodejs builtin `fetch` but a npm library to make http calls.

roschaefer: Oh, I just saw that we use both `node-fetch` *and* `request` npm library. :disappointed:

backend/src/graphql/resolvers/groups.spec.ts
roschaefer: I like this idea. I will try to implement it like this. Ideally we wouldn't need to have two separate functions `contextUser` and `contextConfig` but one.

ulfgebhardt: A separate function forces the developer to make a conscient decision what he wants to mock.

backend/src/config/index.ts
```diff
@@ -37,6 +37,26 @@ const required = {
...
+  conf: typeof required,
+): asserts conf is NonNullableCollection<typeof required> {
+  Object.entries(conf).forEach(([key, value]) => {
+    if (!value) {
+      throw new Error(`ERROR: "${key}" env variable is missing.`)
+    }
+  })
+}
+
+assertRequiredConfig(required)
```
ulfgebhardt: I think this is a regression. Readability has degraded alot.

ulfgebhardt (changes requested):
I, by now believe its no longer viable to merge this PR as it is.

It addresses multiple issues, some of them unrelated and with time the scope increased even further.

Basic corncerns are not addressed while I was requested to review this again (3/4th time?).

Why is `fetch` in the conext?

The readability and maintainability has suffered in some parts of the application
- tests are more complex with the new setup mechanic
- readability has suffered e.g. in the config (required config values)

I suggest - break this up into digestable pieces, ensure we have proper separation of concern within one PR.

ulfgebhardt (approved):

PR #8636, opened 2025-06-04
feat(other): group create feature

Mogge:
## 🍰 Pullrequest
Simple feature to test that the group creation is working.

### Issues

- None

### Todo

- [X] None


mahula (approved):
Good catch

PR #8655, opened 2025-06-09
feat(backend): logger

Mogge:
## 🍰 Pullrequest
Provides a logger object in the context, which can be addressed by `logger.error(...)` or `logger.debug(...)` same as console.

Logs all apollo requests and responses.

### Issues

- None

### Todo

- [ ] Log file rotations on servers

### DIFF
--- a/webapp/components/Chat/Chat.vue
+++ b/webapp/components/Chat/Chat.vue
@@ -69,6 +69,14 @@
           </div>
         </div>
 
+        <div
+          v-for="message in messages.filter((m) => m.isUploading)"
+          :slot="'message_' + message._id"
+          v-bind:key="message._id"
+        >
+          {{ $t('chat.transmitting') }}
+        </div>
+
         <div v-for="room in rooms" :slot="'room-list-avatar_' + room.id" :key="room.id">
           <div
             v-if="room.avatar"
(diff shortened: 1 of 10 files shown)


Bettelstab (approved):

backend/src/server.ts
```diff
@@ -63,6 +78,37 @@ export const context = async (options) => {
...
+export const loggerPlugin = {
+  requestDidStart(requestContext) {
+    const isIntrospectionQuery = requestContext.request.operationName === 'IntrospectionQuery'
+    const qID = randomBytes(4).toString('hex')
+    if (!isIntrospectionQuery) {
+      const logRequest = ['Apollo Request', qID, requestContext.request.operationName]
+      logRequest.push(JSON.stringify(requestContext.request.query))
+      if (requestContext.request.variables) {
+        const variables = cloneDeep(requestContext.request.variables)
+        if (variables.password) variables.password = '***'
```
roschaefer: I would like to see this software-tested before we merge. You can expect `context.logger` to be called with an obfuscated password on mutation `login`.

backend/src/server.ts
```diff
@@ -63,6 +78,37 @@ export const context = async (options) => {
   }
 }
 
+export const loggerPlugin = {
```
roschaefer: Maybe put this into its own file so we can test just the logger plugin with a test server?

roschaefer (changes requested):
familiar

I think I build the very same PR for a different project recently. Was it dreammall?

There is little I have to request changes for but I would like to see a test case for this. Besides this, the PR is pretty much textbook. :+1:

Mogge:
You are right. This is mainly a copy and paste from dreammall

Bettelstab (approved):

PR #8656, opened 2025-06-10
fix(webapp): fix group performance

ulfgebhardt:
## 🍰 Pullrequest

Fix group performance

This fixes the immidiate problem with group performance making the browser crash, by introducing a limit to group members requested. Furthermore the client uses a new component to prevent rerendering of the long scroll list (Thanks @sebastian2357 )

As the search is client side, the search option is now displayed when all members are loaded only.

Hide Serach till all members are loaded
image
image

For testing the seed was extended to seed 3000 users as group members of the yoga group
image

Seed ensures the user slug is available. This results in longer seed time.

The admin list still takes about 15 secs to render
image

### Issues

- None

### Todo

- [X] None


sebastian2357:
I have increased the profileListVisibleCount to 6 in order to fill the free box space 

grafik

sebastian2357 (approved):

PR #8658, opened 2025-06-10
fix(webapp): added timer

sebastian2357:
### Issues

- None

### Todo

- [X] added timer

### DIFF
--- a/backend/src/graphql/resolvers/attachments/attachments.ts
+++ b/backend/src/graphql/resolvers/attachments/attachments.ts
@@ -129,17 +129,22 @@ export const attachments = (config: S3Configured) => {
     }
     const command = new Upload({ client: s3, params })
     const data = await command.done()
-    const { Location } = data
-    if (!Location) {
+    let { Location: location } = data
+    if (!location) {
       throw new Error('File upload did not return `Location`')
+    }
+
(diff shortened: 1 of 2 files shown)


ulfgebhardt (dismissed):

sebastian2357:
> Snapshot testing is broken. 
What exactly was broken? For me the snapshots worked, but I added tests for the delay functionality.

Also, I think the same functionality is already implemented and available as parameter for `openMenu()`. See `Dropdown.vue`.
Yes, the point is just that the implementation there is specially implemented around the popover behind. I prefer the new implementation, because it is simpler and cleaner.

Bettelstab (approved):

Bettelstab:
> > Snapshot testing is broken.
> 
> What exactly was broken? For me the snapshots worked, but I added tests for the delay functionality.
> 
There was a change in the snapshot before which was not supposed to be there - the popup was not in anymore after a hover, because you hadn't wait for it. Now you do and the change is gone, perfect.
(Although possibly it would have been easier with react-testing-library's `waitFor`, but hey.)

> > Also, I think the same functionality is already implemented and available as parameter for `openMenu()`. See `Dropdown.vue`.
> 
> Yes, the point is just that the implementation there is specially implemented around the popover behind. I prefer the new implementation, because it is simpler and cleaner.
Alright.

PR #8688, opened 2025-06-18
refactor(backend): put fetch into context

roschaefer:
It's a best practice to add HTTP API clients to the `context`. Standard Apollo GraphQL tutorial has an example: https://www.apollographql.com/docs/apollo-server/data/context#resolvers

Also database drivers, loggers, message queues or process environment defined configuration should go into the `context`, just like HTTP-clients.

### Issues

- None

### Todo

- [X] None


Mogge (changes requested):
Please see this PR: https://github.com/Ocelot-Social-Community/Ocelot-Social/pull/8640

The request library is removed as it is totally deprecated:
image

Please use the node native fetch instead. Therefore we do not need it in the context

ulfgebhardt (changes requested):
I am still against this change as fetch does not have a state.

Only things with a state should be part of the context.
- The database connection is a state ensuring we do connect only once or with a defined pool
- The config is a global state, not to be modified, but still a state
- Pubsub is debatable but behaves similar as the database and requires a `new` - hence a state

While I find examples where specific APIs are created at application start with `new UserAPI()` [here](https://stackoverflow.com/a/58091977) where the `new` implies a state of this API, I have trouble finding an example with fetch in the context.

Rule of thumb should be: only things that need a state/need to be initialized should be in the context.

A good example would be an API which requires authentication - hence needs to maintain an authenticated state or restore it if authentication ended.

Also the link you [provided](https://www.apollographql.com/docs/apollo-server/data/context#resolvers) has an example where an API is passed to the context. And this makes sense since it has a state (`new` keyword).
Nowhere is stated that this is best practice to pass all APIs to the context.

Especially in the case of stateless/static APIs, I believe it actually ways down the application as
- the context is polluted with stuff not required and passed on with every request
- adds complexity to the application as you require knowledge not only of the API, but where to find it (context)

If you disagree I would love to see some more examples defining this a s best practice or have signification upsides.

PR #8690, opened 2025-06-18
fix(webapp): show hint that message is being saved

Bettelstab:
## 🍰 Pullrequest
So far, there is no indication whatsoever while a chat message is sent to the server. With possible files being uploaded, that can take a while.

I wanted to instantly add the message, but without the checkmark. The last part turned out to be complicated. I guess I could make it disappear with the checkmark slot.

But, to indicate on the image that it is uplpading, there is not so much you can do. I explored possibilities like this
Screenshot 2025-06-18 at 21 25 48
But I don't like it, because the 50% is a fixed value I set and I can only guess the real progress.

My suggestion for now is something like:
Screenshot 2025-06-18 at 21 36 20 (2)

wording tbd.

Unfortunately I don't have a glue how I can make the date disappear.

## TODO
- [x] translate

### DIFF
--- a/backend/src/graphql/resolvers/attachments/attachments.ts
+++ b/backend/src/graphql/resolvers/attachments/attachments.ts
@@ -129,17 +129,22 @@ export const attachments = (config: S3Configured) => {
     }
     const command = new Upload({ client: s3, params })
     const data = await command.done()
-    const { Location } = data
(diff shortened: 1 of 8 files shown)


Tirokk (approved):
Nice!

PR #8714, opened 2025-06-25
fix(backend): refactor S3 usage and always apply protocol fix

Bettelstab:
## 🍰 Pullrequest
Cleanup s3 code, so we use the same code for uploading files in chat and images in posts.

### DIFF
--- a/backend/src/graphql/resolvers/attachments/attachments.ts
+++ b/backend/src/graphql/resolvers/attachments/attachments.ts
@@ -1,13 +1,12 @@
 import path from 'node:path'
 
-import { DeleteObjectCommand, ObjectCannedACL, S3Client } from '@aws-sdk/client-s3'
-import { Upload } from '@aws-sdk/lib-storage'
 import { UserInputError } from 'apollo-server-express'
 import slug from 'slug'
 import { v4 as uuid } from 'uuid'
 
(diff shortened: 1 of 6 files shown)


backend/src/graphql/resolvers/images/imagesS3.spec.ts
```diff
@@ -210,7 +210,8 @@ describe('mergeImage', () => {
           S3_PUBLIC_GATEWAY: 'http://s3-public-gateway.com',
         })
 
-        it('changes the domain of the URL to a server that could e.g. apply image transformations', async () => {
+        // eslint-disable-next-line jest/no-disabled-tests
+        it.skip('changes the domain of the URL to a server that could e.g. apply image transformations', async () => {
```
roschaefer, line 214: That's the only thing which I don't understand, why is it skipped?

roschaefer (changes requested):
great stuff

This is great stuff! I like it a lot.

I'm currently looking into the test implementation. It is very convoluted and obviously needs refactoring. At the time, I tried not to touch the interface of `imageS3.ts` but I think it's necessary now.

roschaefer (approved):
@Bettelstab feel free to merge after merging #8716

PR #8752, opened 2025-07-08
refactor(backend): remove obsolete code

roschaefer:
We kept this code for backwards compatibility but since we already deployed S3 to our kubernetes cluster and we're using it locally, let's remove this code. It will also make it easier to implement the image resize service as it reduces the total amount of code to maintain.


backend/src/graphql/resolvers/messages.ts
```diff
@@ -125,19 +125,17 @@ export default {
 
           const atns: File[] = []
 
-          if (isS3configured(CONFIG)) {
```
roschaefer, line 128: This type-guard was here only to make typescript happy, right @ulfgebhardt?

ulfgebhardt: yes, an assert would be better I guess. There is no option to upload the attachments somewhere else then to s3.

backend/src/config/index.ts
```diff
@@ -169,12 +150,23 @@ const CONFIG = {
   ...neo4j,
   ...sentry,
   ...redis,
-  ...s3,
   ...options,
   ...language,
+  IMAGOR_SECRET,
```
ulfgebhardt: this is dependend on imagor?

ulfgebhardt (commented):
There is portions of imagor in here, will review after merge of that.

backend/src/config/index.ts
roschaefer: Good catch, no it's not of course.

backend/src/graphql/resolvers/messages.ts
roschaefer:
```
  | 'AWS_ACCESS_KEY_ID'
  | 'AWS_SECRET_ACCESS_KEY'
  | 'AWS_ENDPOINT'
  | 'AWS_REGION'
  | 'AWS_BUCKET'
```
are now required in `src/config/index.ts`

ulfgebhardt (approved):
Looks good - think its wise to get rid of the old image mechanism.

PR #8846, opened 2025-08-23
feat(webapp): add location distance in group profile

Tirokk:
## 🍰 Pullrequest

Add location distance in group profile if defined

#### Before

Bildschirmfoto 2025-08-23 um 11 44 55

#### After

Bildschirmfoto 2025-08-23 um 11 46 10

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/webapp/components/LocationInfo/LocationInfo.spec.js
+++ b/webapp/components/LocationInfo/LocationInfo.spec.js
@@ -19,13 +19,27 @@ describe('LocationInfo', () => {
     })
   }
 
-  it('renders with distance', () => {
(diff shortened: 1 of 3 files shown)


webapp/graphql/groups.js
Tirokk, line 1: As @Mogge already tried at the moment we didn't solved the problem that we can't use the `locationFragment`.

ulfgebhardt: fixed

Tirokk: Danke, das habe ich übersehen.

PS: Schau mal @Mogge

PR #8878, opened 2025-09-06
[WIP] fix(backend): fix minio settings

ulfgebhardt:
## 🍰 Pullrequest

fix minio settings - I had errors when running the docker with `docker compose up neo4j mailserver minio minio-mc` and `yarn test` in the backend.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -51,6 +51,7 @@ services:
   minio:
     image: quay.io/minio/minio
     ports:
+      - 9000:9000
       - 9001:9001
     volumes:
       - minio_data:/data
--- a/backend/test/helpers.ts
+++ b/backend/test/helpers.ts
@@ -16,7 +16,7 @@ export const TEST_CONFIG = {
   PRODUCTION_DB_CLEAN_ALLOW: false,
   DISABLED_MIDDLEWARES: [],
   SEND_MAIL: false,
-  PROXY_S3: 'http://minio:9000',
+  PROXY_S3: 'http://localhost:9000',
 
   CLIENT_URI: 'http://webapp:3000',
   GRAPHQL_URI: 'http://localhost:4000',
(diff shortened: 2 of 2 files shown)


docker-compose.override.yml
```diff
@@ -51,6 +51,7 @@ services:
   minio:
     image: quay.io/minio/minio
     ports:
+      - 9000:9000
```
roschaefer: This is not a big deal but `backend` already exposes port 9000, doesn't it? It should crash when you start both minio and backend through docker.

This would be the explanation for your problem: You started `backend` on your host and `minio` via docker.

roschaefer (changes requested):

docker-compose.override.yml
ulfgebhardt: It does not on my end
Without it i get
image

I check again

ulfgebhardt: I have to start the backend aswell - then I can omit the exposure of 9000 - this is not ideal, but a working solution.
I removed the port.

ulfgebhardt: The main issue remains - without this change you cannot run the tests locally without all s3 calls failing.

roschaefer: Hm, this is annoying indeed. The tests might not setup the proxy that forwards the calls to minio. It's really difficult to find a solution that works for all scenarios here. Why don't the tests fail on CI? The backend server is constantly running?

ulfgebhardt: I think because the tests might run in docker?! I am unsure.

roschaefer: Maybe the solution is to not have calls to S3 in our integration tests? Just E2E tests where the backend will be running along with the proxy?

ulfgebhardt: I'd appreciate if you think about a proper solution. We actually have an working s3 local setup. Its fine to test with it. We just have trouble with the service exposure in different environments. Either via backend proxy, via locahost:port or dockeraddress:port.
A solution that works in all scenarios would be very good.

Mogge (approved):

PR #8901, opened 2025-09-14
fix(webapp): fix group images

ulfgebhardt:
## 🍰 Pullrequest

fix group images

before:
image

after:
image

### Issues

- None

### Todo

- [X] None


webapp/graphql/groups.js
```diff
@@ -185,6 +185,9 @@ export const groupQuery = (i18n) => {
         }
         avatar {
           url
+          w320: transform(width: 320)
```
roschaefer, line 188: Maybe it helps to use a GraphQL fragment here so this doesn't happen again (so quickly)?

PR #8902, opened 2025-09-14
refactor(backend): externalize all remaining queries in spec files

ulfgebhardt:
## 🍰 Pullrequest

externalize all remaining queries in spec files

This also fixes several problems with the code:
- neode `Post` model did not enforce `content` properly
- default `orderBy` of posts did not apply when an empty orderBy was provided
- user `activeCategories` did not default to null when undefined 

### Issues

- None

### Todo

- [X] None


backend/src/db/models/Post.ts
```diff
@@ -18,7 +18,7 @@ export default {
   },
   title: { type: 'string', disallow: [null], min: 3 },
   slug: { type: 'string', allow: [null], unique: 'true' },
-  content: { type: 'string', disallow: [null], min: 3 },
+  content: { type: 'string', disallow: [null], required: true, min: 3 },
```
ulfgebhardt, line 21: fix: enforce content

backend/src/middleware/orderByMiddleware.ts
```diff
@@ -6,13 +6,20 @@ import cloneDeep from 'lodash/cloneDeep'
...
+    if (!orderByArg) {
+      fieldNode.arguments.push({
+        kind: 'Argument',
+        name: { kind: 'Name', value: 'orderBy' },
+        value: { kind: 'EnumValue', value: 'sortDate_desc' },
+      })
+    } else if (args.orderBy === undefined) {
+      orderByArg.value = { kind: 'EnumValue', value: 'sortDate_desc' }
+    }
+  }
```
ulfgebhardt, line 22: fix: when orderBy was defined, but undefined, default order did not apply - now it does

backend/src/graphql/resolvers/users.ts
```diff
@@ -638,6 +638,7 @@ export default {
         'allowEmbedIframes',
         'showShoutsPublicly',
         'locale',
+        'activeCategories',
```
ulfgebhardt, line 641: fix: when activeCategories was undefined, result of all user queries were null

backend/src/graphql/resolvers/users.spec.ts
```diff
@@ -196,7 +159,7 @@ describe('UpdateUser', () => {
     })
 
     it('is not allowed to change other user accounts', async () => {
-      const { errors } = await mutate({ mutation: updateUserMutation, variables })
+      const { errors } = await mutate({ mutation: UpdateUser, variables })
```
Mogge: Please as usual.
```js
await expect(mutate({ mutation: UpdateUser, variables })).resolves.toMatchObejct({
  errors: [
    message: 'Not Authorized!',
  ],
})

File: backend/src/graphql/queries/User.ts [hunk 5]
```diff
@@ -1,9 +1,165 @@
...
+      location {
+        distanceToMe
+      }
+      activeCategories
+    }
+  }
+`
+
+export const UserEmailNotificationSettings = gql`
+  query ($id: ID, $name: String, $email: String) {
```
Mogge, line 56: This would be a nice use-case for a fragment. It makes the difference visible on the first glance and reduces the lines of code a lot. Also avoids future bugs if there are new fields and they could be forgotten in some queries,

Mogge (changes requested):
Just two small things. Thanks for doing this!

backend/src/graphql/resolvers/users.spec.ts
ulfgebhardt: Fixed + all others in this file, but out of scope.

backend/src/graphql/queries/User.ts
ulfgebhardt: I want to remove the two additional queries and allow querying the two fields email & emailNotificationSettings. I don't think it makes sense to crash a query with not authorized when querying specific fields. I rather have them nulled.

Mogge (approved):

PR #8933, opened 2025-09-29
feat(webapp): bind local styleguide & fix maintenance page

ulfgebhardt:
## 🍰 Pullrequest

bind local styleguide & fix maintenance page

- Always use local styleguide when building (webapp & maintenance)
- Remove `@human-connection/styleguide`
- Pin nginx version in maintenance docker
- Fix maintenance page
image

Note: if you run the frontend or its tests locally you need to run the following commands beforehand:
```bash
cd styleguide/
yarn install
yarn build:lib
```

### Issues

- based on https://github.com/Ocelot-Social-Community/Ocelot-Social/pull/8932

### How to test

- [ ] Please test if all pages still look as they should, I exchanged the styleguide we include as a package for one we build with the webapp.


Tirokk (approved):
Runs well, if I run `yarn install` in `styleguide` folder.

Cool.

PR #8952, opened 2025-10-06
fix(webapp): fix line-height in styleguide when generated

ulfgebhardt:
## 🍰 Pullrequest

fix line height in styleguide when generated

Background: After including the styleguide the line height is too small. This happens because the units under `webapp/assets/_new/styles/tokens.scss` are wrong.

Old line-height (pre styleguide inclusion):
image

Problematic line-height (after styleguide inclusion):
image

New line-height with this PR:
image

Apparently since we include the styleguide and compile it, we can override values from our code base.

~~**Note**: it is required to rebuild webapp to see this change. Recommended to use `docker compose up webapp --no-deps --build` and after successful build `docker compose up`~~

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -64,7 +64,7 @@ jobs:
           echo "BUILD_COMMIT=${GITHUB_SHA}" >> $GITHUB_ENV
       - run: echo "BUILD_VERSION=${VERSION}-${GITHUB_RUN_NUMBER}" >> $GITHUB_ENV
       #- name: Repository Dispatch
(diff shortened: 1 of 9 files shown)


Tirokk (approved):
That fixes our problem after integration of the styleguide. ;-)

Thanks a lot that you fixed it.

PS: I thought about whether it would be better to have a defined unit that works well here, but realized that no unit means a factor of font size, which is cool with me: <https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#syntax>

PR #8958, opened 2025-10-08
feat(webapp): add reason and call to action on post view page if commenting is disabled

Tirokk:
## 🍰 Pullrequest

Add reason and call to action on post view page if commenting is disabled.

### Issues

- None

### Todo

- [x] rename and refactor `canCommentPost`?
- [x] make a component?
  - [x] I'm not sure if this is really an improvement
  - [ ] may be if we use this component as well for a new feature like:
    - [ ] on visiting a link to a post of a closed group:
      - [ ] show no `404` but explain the reason why one can not see this content and offer the possibility to join the closed group (not for hidden/private groups)
      - [ ] open issue for this feature?
- [ ] move locale from path `settings.blocked-users.explanation.commenting-disabled` to `contribution.comment.commenting-disabled.empty-notice`?

#### Post of a blocked user:

Bildschirmfoto 2025-10-09 um 10 24 09

#### Post of a public group where I'm not a member or a pending member:

Not a member:

I can enter the group and it's reactive. Means that I can comment instantly.

Bildschirmfoto 2025-10-09 um 10 18 00

Pending member:

I can leave the group.

Bildschirmfoto 2025-10-09 um 10 18 51

### DIFF
--- a/webapp/pages/groups/_id/_slug.vue
+++ b/webapp/pages/groups/_id/_slug.vue
@@ -599,7 +599,7 @@ export default {
       error(error) {
         this.$toast.error(error.message)
       },
-      fetchPolicy: 'network-only',
+      fetchPolicy: 'cache-and-network',
     },
     GroupMembers: {
       query() {


webapp/pages/post/_id/_slug/index.vue
```diff
@@ -125,27 +125,78 @@
...
+                      {{
+                        $t('contribution.comment.commenting-disabled.blocked-author.button-label', {
+                          name: post.author.name,
+                        })
+                      }}
+                    </base-button>
+                  </nuxt-link>
+                </ds-space>
+                <!-- no group member -->
+                <ds-space v-else-if="!canCommentPost" centered margin="xxx-small">
```
Tirokk: `canCommentPost` seems a bad naming to me because it's not specific enough, but just to general.
`isBlocked` is not included but disables the commenting as well.

We should find a better name and may be refactor a little bit as well?

PR #8998, opened 2025-10-29
fix(webapp): always link compiled styleguide

ulfgebhardt:
## 🍰 Pullrequest

always link compiled styleguide

### Issues

Fixes the following error occurring in runtime.

branded:
image

plain production:
image

### Repro
1. start the webapp branded (e.g. stage.ocelot.social) `docker compose up webapp --build --no-deps` or on the main repo run `docker compose -f docker-compose.yml up webapp --build --no-deps` to build production mode
2. when running open `http://localhost:3000`
3. The Error appears in docker console when branded and on the webpage when running plain production
4. ...
5. Profit

image

Reason for this behaviour is that the library is not built. The async-validator is used for all forms and a devDependency in the styleguide.

This occurs only in production mode/branded as the  `async-validator` is a dev-dependency in the webapp.
You could add it as dependency, but that is not correct, as the dependency for production is for the styleguide library only.

~~A better~~ The solution ~~would be~~ is to link the proper production build. ~~But this ain't that easy. hence the selected solution.~~

### Todo

- [X] None

### DIFF
--- a/webapp/nuxt.config.js
+++ b/webapp/nuxt.config.js
@@ -257,6 +257,11 @@ export default {
           : `[contenthash:7]${isModern ? '.modern' : ''}_${CONFIG.VERSION}.js`,
       css: ({ isDev }) => (isDev ? '[name].css' : `css/[contenthash:7]_${CONFIG.VERSION}.css`),
     },
+    // babel config
+    babel: {
+      // To prevent  ERROR  [BABEL] Note: The code generator has deoptimised the styling of [..] as it exceeds the max of 500KB.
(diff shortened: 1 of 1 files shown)


mahula (approved):
Good.
This fixes the issue.

Tirokk (changes requested):
I build the stylguide `yarn build` and then compiled the webapp `yarn dev` and I get this error:

Bildschirmfoto 2025-10-31 um 09 23 55

ulfgebhardt:
I fixed the error

PR #9124, opened 2026-01-19
refactor(backend): properly model group-membership

ulfgebhardt:
## 🍰 Pullrequest

properly model group-membership

### Issues

- based on https://github.com/Ocelot-Social-Community/Ocelot-Social/pull/9125

### Todo

- [X] None

### DIFF
--- a/webapp/components/CustomButton/CustomButton.vue
+++ b/webapp/components/CustomButton/CustomButton.vue
@@ -18,7 +18,7 @@
         />
       </base-button>
     </a>
-    <nuxt-link v-else :to="settings.toolTipIdent">
+    <nuxt-link v-else :to="settings.path">
       <base-button
         class="custom-button"
         circle


backend/src/graphql/resolvers/groups.spec.ts
```diff
@@ -1208,16 +1228,28 @@ describe('in mode', () => {
                   data: {
                     GroupMembers: expect.arrayContaining([
                       expect.objectContaining({
-                        id: 'current-user',
-                        myRoleInGroup: 'owner',
+                        user: expect.objectContaining({
+                          id: 'current-user',
+                        }),
+                        membership: expect.objectContaining({
```
Mogge, line 1234: In general: Why `expect.objectContaining`. There is just one field.

ulfgebhardt: There is more then one value in the result if i am correct.

Mogge: Well, in all the queries there is just the role field inside membership. May be there are timestamps in the DB, but we do not query them

PR #9128, opened 2026-01-21
refactor(backend): middleware before/after

ulfgebhardt:
## 🍰 Pullrequest

use dynamic before/after defintion for middleares as suggested by @roschaefer

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/middleware/index.spec.ts
+++ b/backend/src/middleware/index.spec.ts
@@ -95,10 +95,13 @@ describe('default', () => {
         },
         disabledMiddlewares: ['sentry', 'xss'],
       })
+      const consoleSpy = jest.spyOn(console, 'log').mockImplementation()
       const middlewares = getCapturedMiddlewares()
       expect(middlewares).toHaveLength(14)
       expect(middlewares).not.toContain(sentryMarker)
       expect(middlewares).not.toContain(xssMarker)
+      expect(consoleSpy).toHaveBeenCalledWith('Warning: Disabled "sentry, xss" middleware.')
+      consoleSpy.mockRestore()
     })
   })
 })


backend/src/middleware/index.ts
```diff
@@ -24,42 +23,89 @@ import userInteractions from './userInteractions'
...
+  /*
+    - null = push ontop
+    - undefined = append
+    - string = append before/after middleware with name
+  */
+  at?: string | null
+  /*
+    append before or after. Undefined behaves like after.
+  */
+  mode?: 'before' | 'after'
```
roschaefer:
```suggestion
  constraint: 'prepend' | 'append' | { 'before': Middleware } | { 'after': Middleware }
```
```

File: backend/src/middleware/index.ts [hunk 2]
```diff
@@ -24,42 +23,89 @@ import userInteractions from './userInteractions'
...
-  middleware: IMiddleware
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  middleware: IMiddleware | IMiddlewareGenerator<any, any, any>
+}
+
+const ocelotMiddlewares: MiddlewareOrder[] = []
+
+export const addMiddleware = (middleware: MiddlewareOrder) => {
+  switch (middleware.position.at) {
+    case undefined:
```
roschaefer: Prefer specific enumerated types (union types) over unspecific `undefined` or `null` values.

backend/src/middleware/index.ts
```diff
@@ -24,42 +23,89 @@ import userInteractions from './userInteractions'
...
+addMiddleware({ name: 'orderBy', middleware: orderBy, position: { at: 'includedFields' } })
+addMiddleware({ name: 'chatMiddleware', middleware: chatMiddleware, position: { at: 'orderBy' } })
+addMiddleware({ name: 'categories', middleware: categories, position: { at: 'chatMiddleware' } })
 
 export default (schema) => {
-  const middlewares = ocelotMiddlewares
-    .concat(brandingMiddlewares())
-    .sort((a, b) => a.order - b.order)
+  // execute branding middleware function
+  brandingMiddlewares()
```
roschaefer, line 92: Try to avoid side effects. `brandingMiddlewares` apparently mutates `ocelotMiddlewares`. Simply return `ocelotMiddlewares` from your function.

roschaefer (dismissed):
I'd suggest two things:

1. I do remember that there a reason why middlewares had to be removed, added or re-ordered depending on the deployment. There is also the issue of different development cycles, ie. the customization may be slower than ocelot community. However, I would strongly suggest that we check if we can't get rid of this complexity altogether. Especially after we have fewer customers now (less customizations, maybe potential to get rid of this complexity)

2. If we stick to this complexity, let's refactor this PR to be more functional and with stronger typing. See my suggestions below. Feel free to get in touch and we do a pairing.

backend/src/middleware/index.ts
ulfgebhardt: I do not think this is wise as you would requie to import different middlewares for before/after. The branding is not part of the maincode. This actually increases complexity. As developer you just need to know the correct middleware name where you want to plug it in and not the file structure of where to get that Class from. If you make a typo the thing will crash on startup and you know you wrote things wrong. Think its pretty foolproof.

backend/src/middleware/index.ts
ulfgebhardt: Any suggestion when referencing the middleware name string not the IMiddleware Class?

backend/src/middleware/index.ts
ulfgebhardt: I decided for this as I want to use the `addMiddleware` function. This modifies the ocelotMiddlewares. So the side effect stems from there. I need a function to call that method from the branding - thats why I decided for this.

backend/src/middleware/index.ts
ulfgebhardt: I liked the type setup and adjusted the implementation accordingly. I reference a string for middleware to reduce complexity for branding middlewares. You need the name and you are not required to import the corresponding class.

ulfgebhardt:
> Wow. How much AI was involved in this unit test?

100% ai, I just ensured things were correct. I actually found an error that way.

PR #9130, opened 2026-01-23
fix(backend): fix bug in notifications settings for currentUser

ulfgebhardt:
## 🍰 Pullrequest

fix bug in notification settings as the parent did not contain all database values when querying `currentUser`.

Before this patch when logging in the user had all notification settings set to true. Even tho the database has saved `false` as value.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/graphql/resolvers/follow.spec.ts
+++ b/backend/src/graphql/resolvers/follow.spec.ts
@@ -2,21 +2,20 @@
 /* eslint-disable @typescript-eslint/no-unsafe-call */
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
-import { createTestClient } from 'apollo-server-testing'
-
 import Factory, { cleanDatabase } from '@db/factories'
-import { getDriver, getNeode } from '@db/neo4j'
 import { followUser } from '@graphql/queries/followUser'
 import { unfollowUser } from '@graphql/queries/unfollowUser'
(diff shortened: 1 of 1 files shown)


backend/src/graphql/resolvers/user_management.ts
```diff
@@ -18,8 +15,21 @@ const neode = getNeode()
 
 export default {
   Query: {
-    currentUser: async (object, params, context, resolveInfo) =>
-      neo4jgraphql(object, { id: context.user.id }, context, resolveInfo),
+    currentUser: async (_object, _params, context: Context, _resolveInfo) => {
+      if (!context.user) {
```
Mogge, line 19: This can be done before the previous neo4jgraphql call, which makes the code for the active categories obsolete.

ulfgebhardt: I do not understand what you are suggesting

Mogge: Instead of writing a cypher, you can catch the error before calling `neo4jgraphql`. So you do not need the further changes

ulfgebhardt: Thats not the error. The error is, that the Database fields are not resolved if you do not do it manually. Don't ask me why.

ulfgebhardt: `if (!context.user) {` is just a typescript thing and actually is not needed as the query requires authorization.

PR #9155, opened 2026-01-31
refactor(workflow): do not clean cache after run

ulfgebhardt:
## 🍰 Pullrequest

do not clean cache after run

Quite often and especially on, but not limited to `master` we have the following:
image
image

This happens since the github cache was reduced in size. As it seems that the latest cache entries are cleaned automatically and so fast that the cache-cleanup falls short on finding the cache entry, I vote to remove this functionality all together.

### Issues

- None

### Todo

- [X] None


ulfgebhardt (commented):
Geprüft ✅

PR #9241, opened 2026-02-17
refactor(workflow): reuse same cache to prevent double build on webapp

ulfgebhardt:
## 🍰 Pullrequest

reuse same cache to prevent double build on webapp

### Issues

- None

### Todo

- [X] None


ulfgebhardt:
not working

